PR #32, opened 2018-09-19
Review from Dave Cridland

dwd:
* Change reference to XMPP to the current RFC.
* Change heading of "Deniability" section, which also discusses non-repudiation, to "Non-Repudiation and Deniability".

I wondered if that latter section should note that a third-party (ie, non-Member) cannot identify a sender from a given encrypted message, but then I wondered if that was the case or not.


beurdouche:
Thanks @dwd !

PR #38, opened 2018-10-13
requirements language shift

seanturner:
This takes the "de_2119" idea a little bit further by making the requirements less aspirational.  For example instead of saying "MLS should be able to does this cool thing.." we just say "MLS does this cool thing".  Obviously, we will need to ensure that the claims are in fact true before we go final.

Also, by getting rid of lower case should, must, etc. we will avoid the "did you mean to upper case that" kind of questions that are bound to come with requirements documents.

There are also some other editorial nits here.


beurdouche:
Thanks Sean @seanturner !

PR #40, opened 2018-10-14
Use 'MLS Protocol' terminology in the Architecture document.

beurdouche:
Work in progress, please do NOT merge...


beurdouche:
The terminology of this document was adopted instead of the reverse.

PR #41, opened 2018-10-21
Updates based on a review in advance of the November 2018 IETF meeting.

jmillican:
- Remove "join an existing group" from architecture doc, based on UserAdd having been removed from the protocol doc.
- Remove statement that message delivery cannot be correlated between AS identities and DS delivery channels. This seems generally very difficult to achieve. It may be possible at times, but does not seem to be grounded in discussions that we have had at the working group meetings.
- Note that federation requires compatible wire encodings.
- Note that communication with the AS requires TLS or alternative authentication.
- Note that out-of-band channels may facilitate detection of misbehaviour. 
- Some grammar fixes.

### DIFF
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -205,8 +205,6 @@ Clients may wish to do the following:
 
  -  remove one or more members from an existing group;
 
- -  join an existing group;
-
  -  leave a group;
 
  -  send a message to everyone in the group;
@@ -429,16 +427,15 @@ are made about the format of the payload.
 
 The protocol is designed in a way that limits the server-side (AS and DS)
 metadata footprint. The DS only persists data required for the delivery
-of messages and avoid Personally Identifiable Information (PII) or other
-sensitive metadata wherever possible. A Messaging Service provider that has
-control over both the AS and the DS, will not be able to correlate encrypted
-messages forwarded by the DS, with the initial public keys signed by the AS.
+of messages and avoids Personally Identifiable Information (PII) or other
+sensitive metadata wherever possible.
 
 ### Federation
 
 The protocol aims to be compatible with federated environments. While this
 document does not specify all necessary mechanisms required for federation,
-multiple MLS implementations can interoperate and to form federated systems.
+multiple MLS implementations can interoperate and to form federated systems if
+they use compatible wire encodings.
 
 
 ### Compatibility with future versions of MLS
@@ -463,7 +460,8 @@ goals ("MLS aims to guarantee that...")?]]
 We assume that all transport connections are secured via some transport
 layer security mechanism such as TLS {{?I-D.ietf-tls-tls13}}. However,
 as noted above, the security of MLS will generally survive compromise
-of the transport layer.
+of the transport layer. Note that this may not extend to connections with the
+AS, which will typically require authentication at a minimum.
 
 ### Message Secrecy and Authentication {#message-secrecy-authentication}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-architecture.md
```diff
@@ -205,8 +205,6 @@ Clients may wish to do the following:
 
  -  remove one or more members from an existing group;
 
- -  join an existing group;
```
raphaelrobert: I suppose this relates to UserAdd having been removed from the protocol draft?
If so, the idea was to re-introduce it later, so I'm wondering if this requirement shouldn't stay in the architecture draft in the meantime (as a goal for the protocol)

draft-ietf-mls-architecture.md
```diff
@@ -561,7 +559,8 @@ weakening the PCS guarantees for attachments.
 
 In general we do not consider Denial of Service (DoS) resistance to be the responsibility
 of the protocol. However, it should not be possible for anyone to perform a
-trivial DoS attack from which it is hard to recover.
+trivial DoS attack from which it is hard to recover. This excludes the DS, which
```
raphaelrobert: "This excludes the DS" -> "This excludes the DS as an attacker"

draft-ietf-mls-architecture.md
beurdouche: I agree with Raphael, this should be kept.

draft-ietf-mls-architecture.md
```diff
@@ -429,16 +427,15 @@ are made about the format of the payload.
 
 The protocol is designed in a way that limits the server-side (AS and DS)
 metadata footprint. The DS only persists data required for the delivery
-of messages and avoid Personally Identifiable Information (PII) or other
```
beurdouche: Please revert this change, this is not an editorial decision. Moreover I disagree with removing it except if we have consensus that we cannot achieve it.

draft-ietf-mls-architecture.md
```diff
@@ -463,7 +460,8 @@ goals ("MLS aims to guarantee that...")?]]
 We assume that all transport connections are secured via some transport
 layer security mechanism such as TLS {{?I-D.ietf-tls-tls13}}. However,
 as noted above, the security of MLS will generally survive compromise
-of the transport layer.
+of the transport layer. Note that this may not extend to connections with the
+AS, which will typically require authentication at a minimum.
```
beurdouche: What do you mean here ?

draft-ietf-mls-architecture.md
beurdouche: You can revert this change. "Anyone" is already contains the DS.

draft-ietf-mls-architecture.md
```diff
@@ -607,9 +606,9 @@ including total DoS attacks (where it simply refuses to forward any
 messages) and partial DoS attacks (where it refuses to forward
 messages to and from specific Clients). As noted in
 {{delivery-guarantees}}, these attacks are only partially detectable
-by clients. Ultimately, failure of the DS to provide reasonable
-service must be dealt with as a customer service matter, not via
-technology.
+by clients without an out-of-band channel. Ultimately, failure of
```
beurdouche: What do you mean here ?

draft-ietf-mls-architecture.md
jmillican: Oh sorry @beurdouche, I meant that the DS is able to DoS, as opposed to anyone else who shouldn't be able to. Will update to @raphaelrobert's suggestion.

draft-ietf-mls-architecture.md
jmillican: Fair enough :) I wanted to include as a point for discussion, as it seems a fundamentally unrealistic goal to achieve. Let's discuss on the list.

draft-ietf-mls-architecture.md
jmillican: We claim the MLS should be secure even without any sort of transport security. This makes sense for the DS, but if we are to have an AS which authoritatively authenticates identities, this will have to be authenticated.

draft-ietf-mls-architecture.md
jmillican: It feels like too strong an assertion to claim that there can be no technical solution to detecting a targeted DoS by the DS. Something out of band might work; such as a transparency protocol.

PR #53, opened 2019-03-10
Align language and syntax with protocol document

beurdouche:
Mostly lowercase everything and fixes a few misuses of "member" that were to restrictive.

### DIFF
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -233,13 +233,15 @@ devices to interact with the Messaging Service.
 These members will typically correspond to end-user devices such as phones,
 web clients or other devices running MLS, which are called clients.
 
-Each client owns a long term identity key pair that uniquely defines
-its identity to other clients or members a the Group.
+Each client owns at least one long term identity key pair that
+uniquely defines its identity to other clients or members a the Group.
 Because a single user may operate multiple devices simultaneously
 (e.g., a desktop and a phone) or sequentially (e.g., replacing
 one phone with another), the formal definition of a group in MLS
 is the set of clients that has knowledge of the shared group secret
 established in the group key establishment phase of the protocol.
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-architecture.md
```diff
@@ -213,52 +213,52 @@ Clients may wish to do the following:
...
+Each client owns a long term identity key pair that uniquely defines
+its identity to other clients or members a the Group.
+Because a single user may operate multiple devices simultaneously
 (e.g., a desktop and a phone) or sequentially (e.g., replacing
-one phone with another), the formal definition of a Group in MLS
-is the set of Clients that has legitimate knowledge of the shared (Encryption)
-Group Key established in the group key establishment phase of the protocol.
+one phone with another), the formal definition of a group in MLS
+is the set of clients that has knowledge of the shared group secret
+established in the group key establishment phase of the protocol.
```
raphaelrobert: To account for both multi-device modes we discussed, I'd like to add the following:

"Multiple user devices can be grouped, appearing as one virtual client to the rest of the group."

draft-ietf-mls-architecture.md
```diff
@@ -213,52 +213,52 @@ Clients may wish to do the following:
...
-web clients or other devices running MLS.
+Informally, a group is a set of users who possibly use multiple endpoint
+devices to interact with the Messaging Service.
+These members will typically correspond to end-user devices such as phones,
+web clients or other devices running MLS, which are called clients.
 
-Each member device owns a long term identity key pair that uniquely defines
-its identity to other Members of the Group.
-Because a single Member may operate multiple devices simultaneously
+Each client owns a long term identity key pair that uniquely defines
```
raphaelrobert: Since there can be more than one identity key per virtual client, this should be changed to:

"Each client owns at least one long term identity key pair that uniquely defines
its identity to other clients or members a the Group."

PR #55, opened 2019-03-10
The DS ensures UIK identifier uniqueness

beurdouche:
https://github.com/mlswg/mls-protocol/issues/96


beurdouche:
@bifurcation Should I add a sentence in the protocol document as well ?

PR #57, opened 2019-04-19
Removed duplicate word "messages"

nicorsm:
In the section _3.1.7. Compatibility with future versions of MLS_, the word "messages" is duplicated.


beurdouche:
Thanks

PR #58, opened 2019-12-28
Spelling fixes in the multi device section

vanitasvitae:
There were some odd formulations in section 3.1.3.
I hope I addressed those issues without changing the intended meaning.

### DIFF
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -591,7 +591,7 @@ proof is not possible.
 
 MLS adopts the Internet threat model {{?RFC3552}} and therefore assumes
 that the attacker has complete control of the network. It is intended
-to provide the security services described in in the face of such attackers.
+to provide the security services described in the face of such attackers.
 In addition,
 these guarantees are intended to degrade gracefully in the presence
 of compromise of the transport security links as well as of
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -418,11 +418,11 @@ does not result in permanent exclusion from the group.
 
 ### Support for Multiple Devices
 
-It is typically expected for users within Group to own different devices.
+It is typically expected for users within a Group to own different devices.
 
-A new device can be added to a group by sharing of an existing client secrets
+A new device can either be added to a group by sharing secrets of an existing client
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-architecture.md
```diff
@@ -418,11 +418,11 @@ does not result in permanent exclusion from the group.
...
 ### Support for Multiple Devices
 
-It is typically expected for users within Group to own different devices.
+It is typically expected for users within a Group to own different devices.
 
-A new device can be added to a group by sharing of an existing client secrets
+A new device can either be added to a group by sharing secrets of an existing client
 or be considered as a new client by the protocol. This client will not gain access
-to the history even if it is owned by someone who owns another member of the Group.
+to the history even if it is owned by some member of the group who owns another client.
```
beurdouche: Can you please revert that sentence, the new one is incorrect. Users/Someone can own clients and members while a member or a client cannot own another member or client.

vanitasvitae: > A Group using a Messaging Service (MS) comprises a set of participants called Members where each member is typically expected to own multiple devices, called Clients

If I understood this correctly, it doesn't make sense that a member owns another member, right?
Or did I misinterpret the statement above?

vanitasvitae: I noticed, that the following statement contradicts with what I quoted above:

> Informally, a group is a set of *users* who possibly use multiple endpoint devices to interact with the Messaging Service. These *members* will typically correspond to end-user *devices* such as phones, web clients or other devices running MLS, which are called *clients*.

This really confuses me. Are *members* now *users* or *devices*?
You probably want to correct the statement quoted in this posting.

beurdouche: Arf, this is the root cause of the problem... This sentence is incorrect, you can refer to the Protocol document to have a correct terminology. I am about to massively update the Architecture document to fix all of that, but it is not done yet.

The rough idea is that a Client is a set of keys. When it is in a group, it is called a Member. As a User, you can have a lot of Devices and on each devices you may have a lot of groups (hence members) and a lot of potentially unused clients.

My advice is to wait until the interim and focus on the protocol document terminology for now, hopefully I'll have time to update the architecture by then...

PR #67, opened 2020-07-10
Expand privacy section

d1vyank:


chelseakomlo:
I would specify [here](https://github.com/mlswg/mls-architecture/pull/67/files#diff-a09258ea2d61fd11b2530846d6be9395R515) why this privacy property is true- i.e, users sign messages that they send to one another, but messages and signatures are then encrypted under the group key.

PR #78, opened 2021-05-12
Reorganize section 3

cascremers:
Reorganization of Section 3 in preparation of expanding text later.

This PR reorders text for a more logical structure, but does not meaningfully change any text. 
 - Provide separate heading for assumptions.
 - Client/endpoint compromise now precedes compromise of other nodes.
 - Attacks outside of threat model moved to the end.

Joint work with @br-hale @kkohbrok .


beurdouche:
Thanks!

PR #80, opened 2021-05-18
Rewrite AS definition to reflect its role in the Spec

kkohbrok:
In the current definition, the AS is portrayed as a distinct service that could, for example run on a server separate from other MLS participants. However, in the MLS spec, it's more of an abstract layer that fulfills certain functionalities, some of which are likely to run on client devices. For example, an identity key that signs individual group-level signature keys could live in the secure element of a device. From the standpoint of the spec, that key would be part of the AS.

This PR changes the definition of the AS to reflect this, although there are of course multiple ways to tackle this.

The MLS spec is not explicit about the fact that it's the AS that's potentially responsible for message authentication, i.e. to bridge a potential gap between the key used to sign a message and the key in a member's leaf. However, it makes sense that this would be part of an abstract AS layer. If not, I believe it would still make sense to address this in the architecture doc somewhere.

The plan is to add more details on compromise of parts of the abstract AS to the security section in a future PR.

Let us know what you think!

This is joint work with @br-hale and @cascremers.

### DIFF
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -352,10 +352,12 @@ signed by the private key corresponding to a member's Credential, but can
 instead be signed by a different key, which is authenticated by the key in the
 Credential in some way.
 
-While Credential Authentication can be performed by a separate entity, message
+While Credential authentication can be performed by a separate entity, message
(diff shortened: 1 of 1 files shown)


beurdouche:
Thanks !

PR #85, opened 2022-02-28
Edit pass on Abstract, Introduction, and General Setting

bifurcation:
Mainly focused on cleaning up grammar, but also fixing some things that have gone stale, like the idea that the MLS protocol is "abstract".

### DIFF
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -44,40 +44,6 @@ author:
     email: alan@wire.com
 
 informative:
-  MLSPROTO:
-       title: "Messaging Layer Security Protocol"
-       date: 2018
-       author:
-         -  ins: R. Barnes
-            name: Richard Barnes
-            organization: Cisco
-            email: rlb@ipv.sx
-         -
-            ins: B. Beurdouche
-            name: Benjamin Beurdouche
-            organization: Inria
-            email: benjamin.beurdouche@inria.fr
-         -
-            ins: J. Millican
-            name: Jon Millican
-            organization: Facebook
-            email: jmillican@fb.com
-         -
-            ins: E. Omara
-            name: Emad Omara
-            organization: Google
-            email: emadomara@google.com
-         -
-            ins: K. Cohn-Gordon
-            name: Katriel Cohn-Gordon
-            organization: University of Oxford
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-architecture.md
```diff
@@ -296,13 +251,13 @@ public values such as a name (an identity), a public encryption key
...
+to other clients or members in the Group.
 In some messaging systems, clients belonging to the same user must
-all share the same identity key pair, but MLS does not assume this.
+all share the same signature key pair, but MLS does not assume this.
 
-Users will typically own multiple clients, potentially one or more per
-end-user devices (phones, web clients or other devices...) and may
+Users will often use multiple clients, potentially one or more per
+device (phones, web clients or other devices...), and may
 choose to authenticate using the same signature key across devices,
```
kkohbrok, line 261: While you're at it, I believe it's not really possible anymore to use the same signature key across client. At least not if two client are meant to be members of the same group. The application could choose to authenticate the credentials based on some other key material, but that really depends on how the AS is designed. I would suggest we discuss the different options in the section describing the AS instead.

kkohbrok (commented):
Looks good! Just two comments on client identity.

draft-ietf-mls-architecture.md
```diff
@@ -296,13 +251,13 @@ public values such as a name (an identity), a public encryption key
 and a public signature key. Ownership of a client by a user is
 determined by the fact that the user has knowledge of the
 associated secret values. When a client is part of a Group, it is
-called a Member and its signature key pair uniquely defines its
-identity to other clients or members in the Group.
+called a Member and its signature key pair uniquely identifies it
+to other clients or members in the Group.
```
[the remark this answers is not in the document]
bifurcation: Yeah, but I don't think we want to get into that detail here.  The fault here is probably in focusing on identity as opposed to saying more directly that the key pair is used to authenticate messages from that client.  I'll try to reword.

draft-ietf-mls-architecture.md
bifurcation: Fair point, will try to reword.

beurdouche:
This looks good ! Thank you !

PR #86, opened 2022-03-01
Update Authentication Service description

bifurcation:
This PR is a little more invasive than the last few, because the Authentication Service section still had a few ideas that may have been current early in the WG's life, but are no longer true.  In particular:

* Message authentication is done with the credential key pair.  Otherwise it's basically impossible to have a self-contained MLS stack, since you would have a gap between the credential key pair and the message signing key pair.

* There is no notion of hierarchically signed keys in MLS, except in the sense of roll-over from one epoch to another.

In addition to fixing these issues, I added some examples to help map the abstract concept to concrete things people might be familiar with.

### DIFF
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -341,7 +341,7 @@ examples illustrate the breadth of this concept:
   public key, with no information to assist in verification).  The verification
   function is the application functionality that enables users to verify keys.
 
-* In a system based on Key Transparency (KT) {{?KeyTransparency}}, the issuance
+* In a system based on Key Transparency (KT) {{KeyTransparency}}, the issuance
   function would correspond to the insertion of a key in a KT log under a user's
   identity. The verification function would correspond to verifying a key's
   inclusion in the log for a claimed identity, together with the KT log's
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -317,14 +317,36 @@ members cannot assume that the Member is a member of the group.
 
 The Authentication Service (AS) has to provide two functionalities:
 
-1. authenticate the credentials (i.e. the identity/signature keypair)
-   used in a group
-2. authenticate messages sent in groups given the signature over the
-   message and the sending member's credential
+1. Issue credentials to clients that attest to bindings between identities and
+   signature key pairs
+
+2. Enable a group member to verify that a credential presented by another member
+   is valid
+
+A member with a valid credential authenticates its MLS messages by signing them
+with the private key corresponding to the public key in its credential.
 
 The AS is considered an abstract layer by the MLS specification, part
 of this service could be, for instance, running on the members'
-devices, while another part is a separate entity entirely.
+devices, while another part is a separate entity entirely.  The following
+examples illustrate the breadth of this concept:
+
+* A PKI could be used as an AS {{?RFC5280}}.  The issuance function would be
+  provided by the certificate authorities in the PKI, and the verification
+  function would correspond to certificate verification by clients.
+
+* Several current messaging applications rely on users verifying each others'
+  key fingerprints for authentication.  In this scenario, the issuance function
+  is simply the generation of a key pair (i.e., credential is just an identifier and
+  public key, with no information to assist in verification).  The verification
+  function is the application functionality that enables users to verify keys.
+
+* In a system based on Key Transparency (KT) {{?KeyTransparency}}, the issuance
+  function would correspond to the insertion of a key in a KT log under a user's
+  identity. The verification function would correspond to verifying a key's
+  inclusion in the log for a claimed identity, together with the KT log's
+  mechanisms for a user to monitor and control which keys are associated to
+  their identity.
 
 By the nature of its roles in MLS authentication, the AS is invested
 with a large amount of trust and the compromise of one of its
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-architecture.md
```diff
@@ -333,39 +355,13 @@ impersonate group members. We discuss security considerations
...
-member's credential, but the key must be authenticated using some
-mechanism. Thus, message authentication relies on the accuracy of
-the key's authentication vice the credential authentication.
-
-While credential authentication can be performed by a separate entity,
-message authentication should be performed by each member separately
-due to the encryption layer of the protocol which protects the
-signature of the message.
+The association between members' identities and signature keys is fairly
+flexible in MLS.  As noted above, there is no requirement that all appearances
```
kkohbrok, line 359:
```suggestion
flexible in MLS.  As noted above, there is no requirement that all clients
```

kkohbrok: As noted in #85, the spec doesn't allow this as long as each client is represented by its own leaf.

draft-ietf-mls-architecture.md
```diff
@@ -317,14 +317,36 @@ members cannot assume that the Member is a member of the group.
...
-2. authenticate messages sent in groups given the signature over the
-   message and the sending member's credential
+1. Issue credentials to clients that attest to bindings between identities and
+   signature key pairs
+
+2. Enable a group member to verify that a credential presented by another member
+   is valid
+
+A member with a valid credential authenticates its MLS messages by signing them
+with the private key corresponding to the public key in its credential.
```
kkohbrok, line 327: This does not reflect what we currently have in the spec. I believe we left this up to the AS/application to allow a deniable key exchange OOB, where messages are then signed by the deniable keys. My guess is that most applications will decide to do it as you describe here. At the same time, I think leaving the option for a gap to allow for a deniable key exchange provides some value at almost no cost.

draft-ietf-mls-architecture.md
kkohbrok, line 359: Maybe add here that it's flexible, because pretty much everything is up to the AS to decide.

bifurcation: In principle, it is possible, as long as the two clients share their entire cryptographic state.  Obviously, that's a lot more challenging than just sync'ing a signature key, but might still be feasible?

draft-ietf-mls-architecture.md
bifurcation: I'm not sure how you come to that conclusion from reading the spec.  It looks to me like the spec is very explicit about exactly which keys are used:

[KeyPackage](https://github.com/mlswg/mls-protocol/blob/main/draft-ietf-mls-protocol.md#keypackage-validation)

> Verify that the signature on the KeyPackage is valid using the public key in the LeafNode's credential

[Commit](https://github.com/mlswg/mls-protocol/blob/main/draft-ietf-mls-protocol.md#processing-a-commit)

> In particular, the signature on the enclosing MLSPlaintext MUST verify using the public key for the credential in the leaf_node of the path field.
> External Commits MUST be signed by the new member. In particular, the signature on the enclosing MLSPlaintext MUST verify using the public key for the credential in the leaf_node of the path field.

[GroupInfo](https://github.com/mlswg/mls-protocol/blob/main/draft-ietf-mls-protocol.md#joining-via-welcome-message)

> Verify the signature on the GroupInfo object. The signature input comprises all of the fields in the GroupInfo object except the signature field. The public key and algorithm are taken from the credential in the leaf node of the member with LeafNodeRef signer. 

[MLSPlaintext / MLSCiphertext in general](https://github.com/mlswg/mls-protocol/blob/main/draft-ietf-mls-protocol.md#content-authentication)

> The signature field in an MLSMessageAuth object is computed using the signing private key corresponding to the public key, which was authenticated by the credential at the leaf of the tree indicated by the sender field

draft-ietf-mls-architecture.md
bifurcation: Well, I meant that it's flexible in the senses described here: That a single user can have multiple key pairs, and a single client's key pairs can change over time.

draft-ietf-mls-architecture.md
bifurcation: That said, my understanding of the deniability situation is that the MLS messages are deniable if the AS is deniable.  Suppose the AS authenticates keys deniably, in the sense of not proving to a third party that the relying client didn't make up the authentication proof themself (one could imagine, e.g., an integration with X3DH).  The only thing the MLS signatures prove is that the holder of the private key sent a message.  So if the AS makes the records of who owns which private key falsifiable, then the whole thing is falsifiable.

So there's no need for the layer of ambiguity that you mention.

bifurcation: @raphaelrobert @kkohbrok - Can you point me to where in the spec you see this notion that messages can be signed by anything other than the key in the credential?  Because the text cited above says the opposite.

draft-ietf-mls-architecture.md
kkohbrok: Then, from an MLS standpoint, it's really the same client 🤷🏻‍♂️.

draft-ietf-mls-architecture.md
kkohbrok: In Section 7.1, the spec says:
> The signature field in an MLSMessageAuth object is computed using the signing private key corresponding to the public key, which was authenticated by the credential at the leaf of the tree indicated by the sender field.

I agree that other places in the spec seem to be more explicit about this. In general, it would probably be best if we had only one place that deals with the verification of signatures on an MLSMessage. If we want to keep the option for deniability, we'd have to include this possible level of indirection. It's optional, so people can (and will in most cases) decide to use the signature key in the credential. Also, it really doesn't cost much, except maybe in implementation complexity, but we can put a MAY there, so implementers can choose to allow this or not.

draft-ietf-mls-architecture.md
kkohbrok: Fair enough.

kkohbrok (approved):

draft-ietf-mls-architecture.md
raphaelrobert: This is now resolved with the latest changes in the protocol spec.

beurdouche:
**Merged** manually in `main` due to rebase conflicts.

PR #87, opened 2022-03-08
Revise the DS section

bifurcation:
This edit got a little invasive, largely around the "ordering" requirements.  As was already pointed out the epoch numbering provides ordering to first order.  What is really needed is for the DS to select a single Commit per epoch.

### DIFF
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -478,7 +478,7 @@ both of the following examples provide a unique Commit per epoch:
 * A "filtering server" Delivery Service where a server rejects all but the
   first Commit for an epoch and clients apply each Commit they receive.
 
-* An "ordered server" Delivery Service where a server forwards all messages
+* An "ordering server" Delivery Service where a server forwards all messages
   but assures that all clients see Commits in the same order, and clients.
 
 * A "passive server" Delivery Service where a server forwards all messages
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -475,7 +475,7 @@ server and client components. Acheiving required uniqueness property will
 typically require a combination of client and server behaviors.  For example,
 both of the following examples provide a unique Commit per epoch:
 
-* An "filtering server" Delivery Service where a server rejects all but the
+* A "filtering server" Delivery Service where a server rejects all but the
   first Commit for an epoch and clients apply each Commit they receive.
 
 * An "ordered server" Delivery Service where a server forwards all messages
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -473,7 +473,7 @@ Service must transmit to the group exactly one Commit message per epoch.
 Much like the Authentication Service, the Delivery Service can be split between
 server and client components. Acheiving required uniqueness property will
 typically require a combination of client and server behaviors.  For example,
-both of the following examples provide a unique Commit per epoch:
+all of the following examples provide a unique Commit per epoch:
 
 * A "filtering server" Delivery Service where a server rejects all but the
   first Commit for an epoch and clients apply each Commit they receive.
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -86,6 +86,26 @@ informative:
        -
           ins: Google
 
+  Loopix:
+      title: "The Loopix Anonymity System"
+      date: 2017
+      author:
+        -
+          ins: A.M. Piotrowska
+          name: Ania M. Piotrowska 
+        -
+          ins: J. Hayes
+          name: Jamie Hayes
+        -
+          ins: T. Elahi
+          name: Tariq Elahi
+        -
+          ins: S. Meiser
+          name: Sebastian Meiser
+        -
+          ins: G. Danezis
+          name: George Danezis
+
 
 --- abstract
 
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-architecture.md
```diff
@@ -410,8 +422,10 @@ cryptographic information:
 * A credential from the Authentication Service attesting to the
   binding between the identity and the client's signature key.
 
-* The client's asymmetric encryption public key material signed with
-  the signature public key associated with the credential.
+* The client's asymmetric encryption public key material.
+
+All the parameters in the KeyPackage are signed with the signature private key
+corresponding to to the credential.
```
kkohbrok, line 428:
```suggestion
corresponding to the credential.
```

draft-ietf-mls-architecture.md
```diff
@@ -431,46 +445,66 @@ the group.
...
+messages to apply.  There are a variety of ways to achieve this agreement, but
+most of them rely on some help from the Delivery Service.  For example, if a
+Delivery Service provides delivery in the same order to all group members, then
+the members can simply apply Commits in the order in which they appear.
+
+For the most part, MLS does not require the Delivery Service to deliver messages
+in any particular order.  The one requirement is that because an MLS group has a
+linear history, the members of the group must agree on the order in which
+changes are applied.  Concretely, the group must agree on which MLS Commit
+messages to apply.
```
kkohbrok, line 468: I think the above paragraph is a dupe.

draft-ietf-mls-architecture.md
```diff
@@ -431,46 +445,66 @@ the group.
...
+delivered to all members.  The Delivery Service may enable these delivery
+patterns via unicast channels (sometimes known as "client fanout"), broadcast
+channels ("server fanout"), or a mix of both.
+
+For the most part, MLS does not require the Delivery Service to deliver messages
+in any particular order.  The one requirement is that because an MLS group has a
+linear history, the members of the group must agree on the order in which
+changes are applied.  Concretely, the group must agree on which MLS Commit
+messages to apply.  There are a variety of ways to achieve this agreement, but
+most of them rely on some help from the Delivery Service.  For example, if a
```
kkohbrok, line 460: If we decide to cast the DS as an abstract layer rather than a service (as done below), we could mandate that the DS does actually have to deal with the problem of deciding which commit to merge.

draft-ietf-mls-architecture.md
```diff
@@ -431,46 +445,66 @@ the group.
...
+in any particular order.  The one requirement is that because an MLS group has a
+linear history, the members of the group must agree on the order in which
+changes are applied.  Concretely, the group must agree on which MLS Commit
+messages to apply.
+
+Each Commit is premised on a given state or "epoch" of the group.  The Delivery
+Service must transmit to the group exactly one Commit message per epoch.
+
+Much like the Authentication Service, the Delivery Service can be split between
+server and client components. Acheiving required uniqueness property will
```
kkohbrok, line 474:
```suggestion
server and client components. Achieving the required uniqueness property will
```

It might make sense to lead with the statement that the DS can be split between client and server. Right now, the paragraphs above here read like it's describing a service running on a server.

draft-ietf-mls-architecture.md
```diff
@@ -431,46 +445,66 @@ the group.
...
+typically require a combination of client and server behaviors.  For example,
+both of the following examples provide a unique Commit per epoch:
+
+* An "filtering server" Delivery Service where a server rejects all but the
```
kkohbrok, line 478:
```suggestion
* A "filtering server" Delivery Service where a server rejects all but the
```

draft-ietf-mls-architecture.md
```diff
@@ -431,46 +445,66 @@ the group.
...
+changes are applied.  Concretely, the group must agree on which MLS Commit
+messages to apply.
+
+Each Commit is premised on a given state or "epoch" of the group.  The Delivery
+Service must transmit to the group exactly one Commit message per epoch.
+
+Much like the Authentication Service, the Delivery Service can be split between
+server and client components. Acheiving required uniqueness property will
+typically require a combination of client and server behaviors.  For example,
+both of the following examples provide a unique Commit per epoch:
```
kkohbrok, line 476:
```suggestion
all of the following examples provide a unique Commit per epoch:
```

draft-ietf-mls-architecture.md
```diff
@@ -431,46 +445,66 @@ the group.
...
+
+* An "filtering server" Delivery Service where a server rejects all but the
+  first Commit for an epoch and clients apply each Commit they receive.
+
+* An "ordered server" Delivery Service where a server forwards all messages
```
kkohbrok, line 481:
```suggestion
* An "ordering server" Delivery Service where a server forwards all messages
```

draft-ietf-mls-architecture.md
```diff
@@ -431,46 +445,66 @@ the group.
...
+  but assures that all clients see Commits in the same order, and clients.
+
+* A "passive server" Delivery Service where a servier forwards all messages
```
kkohbrok, line 484:
```suggestion
* A "passive server" Delivery Service where a server forwards all messages
```

draft-ietf-mls-architecture.md
```diff
@@ -492,10 +526,9 @@ clients. In addition, there may be applications of MLS in which the group
 membership list is stored on some server associated with the Delivery
 Service.
 
-While this knowledge is not a break of authentication or
-confidentiality, it is a serious issue for privacy. In the case where
-metadata has to be persisted for functionality, it SHOULD be stored
-encrypted at rest.
+While this knowledge is not a breach of the protocol's authentication or
+confidentiality guarantees, it is a serious issue for privacy.  Applications
+should consider anonymous systems for server fanout such as Loopix {{Loopix}}.
```
kkohbrok, line 531: I thought this paragraph was w.r.t. the membership list, so encryption at rest makes sense. Maybe include the link to Loopix in the sentence that mentions traffic analysis?

kkohbrok (commented):
Thanks for the PR! Left a few comments, but otherwise looks good!

draft-ietf-mls-architecture.md
beurdouche: I agree.

draft-ietf-mls-architecture.md
beurdouche: Will re-introduce

PR #88, opened 2022-03-10
Revise functional requirements

bifurcation:


draft-ietf-mls-architecture.md
```diff
@@ -622,8 +611,9 @@ use them for tracking.
 > Prefer using encrypted group operation messages to avoid privacy
 > issues related to non-encrypted signatures.
 
-Note that in the default case of encrypted handshake messages, the
-application level must make sure that the access control policies are
+Note that in the default case of encrypted handshake messages, any access
+control policies will be applied at the clien, so the
```
kkohbrok, line 615:
```suggestion
control policies will be applied at the client, so the
```
Maybe mention here that an extension is a great way of making sure all members agree on something? 🙂

draft-ietf-mls-architecture.md
```diff
@@ -635,20 +625,17 @@ states, breaking their ability to communicate.
 
 ### Recovery After State Loss
 
-Group members whose local MLS state is lost or corrupted
-can reinitialize their state and continue participating in the
-group. This does not provide the member with access to group
+Group members whose local MLS state is lost or corrupted can reinitialize their
+state by re-joining the group as a new member and removing the member
+representing their earlier state.  An application can require that a client
+performing such a reinitialization proof its prior membership with a PSK.
```
kkohbrok, line 631:
```suggestion
performing such a reinitialization prove its prior membership with a PSK.
```

kkohbrok, line 631: Maybe we should note here that such a mechanism can allow an adversary to circumvent PCS. For example, a client gets compromised and recovers via an update. Then the adversary decides that they still want access to the group and does a re-join. One way to mitigate this is for clients to require that the PSK presented by a re-joining client not be older than the epoch resulting from the re-joining client's last update.

draft-ietf-mls-architecture.md
```diff
@@ -657,14 +644,22 @@ a new client by the protocol. This client will not gain access to the history
...
 applications can elect to provide such a mechanism outside of MLS.
-Such mechanisms, if used, may undermine the FS and PCS guarantees
+Such mechanisms, if used, may reduce the FS and PCS guarantees
 provided by MLS.
 
-### Extensibility / Pluggability
+### Extensibility
+
+The MLS protocol provides several extension points where additional information
+can be provided.  Extensions to KeyPackages allow clients to disclose additional
```
kkohbrok, line 653: It's not clear to me what this first sentence means. Additional information provided where/to whom? Maybe it would make sense to structurally introduce the different extension categories here, i.e. one bullet point per extension category.

draft-ietf-mls-architecture.md
beurdouche: @kkohbrok I think this should go in the Protocol document, could you add a PR please?

draft-ietf-mls-architecture.md
beurdouche: Looks fine, I think this might be a language thing.

Issue #92, opened 2022-03-23
define a compact-but-exhaustive representation of parameters needed for a concrete deployment of MLS


dkg:
When deploying MLS, the deployer/implementer needs a certain set of choices made.  For example, authentication authorities, transport choices, how to interpret application data, etc.

This draft could create a succinct summary format that a deployer could use as a checklist to make sure they've done the legwork necessary to be able to deploy MLS safely.

this came up in the IETF 113 meeting.

dkg:
If the draft describes such a summary as a template, then maybe some of the existing deployments could try filling it in, so we could see whether it's sufficient.

bifurcation:
I actually think the answer here is fairly compact, and mostly laid out in the arch spec, so it's mainly a question of pulling together the threads.  Basically, the application needs to define:

1. An authentication service
2. A delivery service
3. What the content of the messages will be

dkg:
compact is good ☺  and it looks like your list aligns roughly with the intuitions i opened this issue with, so that's good too.

Are there any choices about cryptographic primitives that need to be decided as well (e.g. ciphersuite limitations or requirements, expected credential types)?  What about MLS extensions or MLS proposals that clients connected to this deployment are expected to support or avoid?

dkg:
I have no idea whether the template i outlined in #93 is correct, let alone exhuastive.  I offer it as a starting point, and won't be upset if anyone wants to step in to improve/correct it.

PR #93, opened 2022-03-29
Add a deployment template

dkg:
The idea here is that a group wanting to deploy MLS can enumerate
their decisions clearly by filling in the template.

It should be possible to ask the developers of a specific MLS
deployment to describe their choices, so that auditors/evaluators can
compare them.

Closes: #92


beurdouche:
I’ll take a stab at completing this soon.

bifurcation:
Interim 2022-06-09:

* Closing in favor of #104

PR #96, opened 2022-04-07
Convert ascii graph to aasvg.

Bren2010:


bifurcation (approved):
Just so folks are aware, here is how the figure renders now:

image

beurdouche:
Thank you Brendan!

PR #98, opened 2022-05-09
Add references to academic work

kkohbrok:
This PR adds a short paragraph referencing the various academic works that has gone into MLS. 

The references to the various works still need to be cleaned up, which I will do once the list is complete. In particular, each reference needs to be checked for correctness (eprint version vs. publicized version) and then included as informal reference in the yaml header of the document.

### DIFF
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -1376,6 +1376,32 @@ specification, which remains non-trivial.
 > clients from physical compromise. In such setting, HSMs and secure
 > enclaves can be used to protect signature keys.
 
+## Cryptographic Analysis of the MLS Protocol
+
+Various academic works have analyzed MLS and the different security guarantees
+it aims to provide. The security of large parts of the protocol has been
+analyzed by [BBN19] (draft 7), [ACDT21] (draft 11) and [AJM20] (draft 12).
+
+Individual components of various drafts of the MLS protocol have been analyzed in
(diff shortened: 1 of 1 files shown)


Bren2010:
- Is there another RFC you can point to that links academic papers on the protocol? I'd expect TLS 1.3 to be a good example of this, but I don't see them do this.
- The references should be cited properly if we do include it

ekr:
> Is there another RFC you can point to that links academic papers on the protocol? I'd expect TLS 1.3 to be a good example of this, but I don't see them do this.

The references are in the appendices. For instance

The reader should refer to the following references for analysis of

https://datatracker.ietf.org/doc/html/rfc8446#appendix-E.1.6

   The reader should refer to the following references for analysis of
   the TLS handshake: [DFGS15], [CHSV16], [DFGS16], [KW16], [Kraw16],
   [FGSW16], [LXZFH16], [FG17], and [BBK17].

kkohbrok:
TLS 1.3 as cited by @ekr was indeed the inspiration for this. The idea was to just have a small paragraph to point the reader in the right direction and then have the full bibliographic entry in the "Informal References" section. As noted above, I intend to properly format all the references once we agree on what references to include, as formatting them into the right yaml is a bit of a pain.

bifurcation (approved):

beurdouche:
Thanks Konrad!

PR #100, opened 2022-05-13
Clarify that AS and DS can be realized in several ways

### LINKED ISSUE #99: AS / DS Terminology
This is a discussion point for the WG, regarding choice of terminology for the AS / DS. For those within the WG, these terms are clear - there have been several issues/PRs to enhance clarity or functionalities (#92, #75, #62, #54, #44, etc.), and the AS / DS describe core functionalities. These are not server components but conceptual representations, e.g. those functionalities could be realized through a "separate directory functionality or service" (spec).

However, I have noticed a very strong trend for people reading the documents / determining use-case applicability to interpret the AS / DS "services" as an absolute requirement for centralized "servers". This does not seem to be a set of random side cases; there is a natural association of terminology and that could result in unnecessary avoidance of the protocol under such a restricted interpretation.

I suggest that we consider a slight terminology change that could avoid these issues. For example, some options may be AM / DM (Authentication Management / Distribution Management), AP / DM (Authentication Provisioning, Distribution Management). I.e. 'management', 'provider / provisioning', 'function', etc. avoid the key word association to 'server' and leave the reader assumptions open to other options.

bifurcation:
Fixes #99

### DIFF
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -164,7 +164,19 @@ clients to prepare for sending and receiving messages securely:
 
 For convenience, this document adopts the representation of these
 services being standalone servers, however the MLS protocol design is
-made so that it is not necessarily the case.
+made so that it is not necessarily the case.  These services may reside on the
+same server or different servers; they may be distributed between server and
+client components; and they may even involve some action by users.  For example:
+
+* Several secure messaging services today provide a centralized DS, and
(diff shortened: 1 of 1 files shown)


br-hale:
That looks good. Thank you for drafting it.

beurdouche (approved):

beurdouche:
Thanks Richard!

PR #101, opened 2022-05-13
Clarifications around PSKs and external operations

### LINKED ISSUE #95: make document self-consistent about PSKs, External Proposals/Commits
The architecture document mentions PSKs in several places, but does not discuss the security of them. Nor does it discuss the security of External Proposals and External Commits.  Notably, the third paragraph of 5.1 Membership Changes says "…the set of devices controlled by the user can only be altered by an *authorized member of the group*."  The set of devices could be altered via an External Commit by a non-member.

At the end of 5.2 Parallel Groups there is a vague mention of "using the PSK mechanism to link healing properties among parallel groups.", but no discussion of the security. Likewise Section 5.5 Recovery After State Loss talks about PSKs as the solution to a problem but the risks are not discussed.

External Proposals and External Commits should be discussed in section 5.4 Access Control.

bifurcation:
Fixes #95

### DIFF
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -514,7 +514,7 @@ subset thereof.  Such policies can be implemented at the application layer, on
 top of MLS. Regardless, MLS does not allow for or support addition
 or removal of group members without informing all other members.
 
-Membership of an MLS group is the managed at the level of individual clients.
+Membership of an MLS group is managed at the level of individual clients.
 In most cases, a client corresponds to a specific device used by a user. If a
 user has multiple devices, the user will be represented in a group by multiple
 clients.  If an application wishes to implement operations at the level of
@@ -572,7 +572,7 @@ membership.
 Applications may use the PSK mechanism to link healing properties among parallel
 groups.  For example, suppose a common member M of two groups A and B has
 performed a key update in group A but not in group B.  The key update provides
-PCS with regard to M in group A.  If a PSK exported from group A and injected
+PCS with regard to M in group A.  If a PSK is exported from group A and injected
 into group B, then some of these PCS properties carry over to group B, since the
 PSK and secrets derived from it are only known to the new, updated version of M,
 not to the old, possibly compromised version of M.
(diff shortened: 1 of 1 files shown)


rohan-wire (approved):

draft-ietf-mls-architecture.md
```diff
@@ -604,13 +627,35 @@ states, breaking their ability to communicate.
...
 state by re-joining the group as a new member and removing the member
 representing their earlier state.  An application can require that a client
 performing such a reinitialization prove its prior membership with a PSK.
 
+There are a few costs to this approach.  If a group uses such PSKs as
+the sole form of authorization to join the group, there is a risk that leaked
+PSKs could allow an unauthorized party to join.  At a practical level, the
+application will need to ensure that the new members have the required PSK,
+including any new members that have joined the group since the epoch in which
+the PSK was issued.
```
Bren2010, line 657: The second sentence seems like a red herring since an external add-remove combo checks that the new leaf has an equivalent credential as the old leaf. Yes you *could* do what you say, but the first paragraph indicates that we're not. I would just keep the third sentence, which is a very good point.

beurdouche:
Thanks both, I'll have a look today.

draft-ietf-mls-architecture.md
```diff
@@ -604,13 +627,28 @@ states, breaking their ability to communicate.
...
+
 ## Recovery After State Loss
 
 Group members whose local MLS state is lost or corrupted can reinitialize their
 state by re-joining the group as a new member and removing the member
 representing their earlier state.  An application can require that a client
 performing such a reinitialization prove its prior membership with a PSK.
 
+There are a few practical challenges to this approach.  For example, the
+application will need to ensure that the new members have the required PSK,
```
Bren2010, line 648:
```suggestion
application will need to ensure that all members have the required PSK,
```

Bren2010 (approved):

beurdouche:
Looks good ! Thanks Richard!

PR #104, opened 2022-06-02
Add a section on interoperability requirements.

### LINKED ISSUE #92: define a compact-but-exhaustive representation of parameters needed for a concrete deployment of MLS

Bren2010:
Replaces #93 
Closes #92 
Not compact.

### DIFF
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -379,7 +379,7 @@ view of the security analysis.
 Upon joining the system, each client stores its initial cryptographic
 key material with the Delivery Service. This key material, called a
 KeyPackage, advertises the functional abilities of the client such as
-supported protocol versions, extensions, and the following
+supported protocol versions, supported extensions, and the following
 cryptographic information:
 
 * A credential from the Authentication Service attesting to the
@@ -696,27 +696,33 @@ software is updated while the client is a member of a group. Thus, in addition
 to preventing downgrade attacks, the members of a group can also observe when it
 is safe to upgrade to a new ciphersuite or protocol version.
 
-# Interoperability Requirements
+# Operational Requirements
 
-A fully functional deployment of MLS requires a number of infrastructure
-services to be available, as well as several policies and design trade-offs that
-are decided by the application. This section lists all of the dependencies of an
-MLS deployment that are external to the protocol specification, but would still
-need to be aligned for two deployments to potentially interoperate.
+MLS is a security layer that needs to be integrated with an application. A
+fully-functional deployment of MLS will have to make a number of decisions about
+how MLS is configured and operated.  Deployments that wish to interoperate will
+need to make compatible decisions. This section lists all of the dependencies of
+an MLS deployment that are external to the protocol specification, but would
+still need to be aligned within a given MLS deployment, or for two deployments
+to potentially interoperate.
 
-MLS has the built-in ability to negotiate protocol versions, ciphersuites,
-extensions, and credential types. For two deployments to interoperate, they must
-have overlapping support in each of these categories.
+The protocol has a built-in ability to negotiate protocol versions,
+ciphersuites, extensions, credential types, and additional proposal types. For
+two deployments to interoperate, they must have overlapping support in each of
+these categories. A `required_capabilities` extension can help maintain
+interoperability with a wider set of clients by ensuring that certain
+functionality continues to be supported by a group, even if the clients in the
+group aren't currently relying on it.
 
 MLS relies on the following network services. These network services would
-either need to be the same, or compatible at the wire-format level, in order for
+need to be compatible in order for
 two different deployments based on them to interoperate.
 
 - An **Authentication Service**, described fully in {{authentication-service}},
   defines the types of credentials which may be used in a deployment and
   provides methods for:
   1. Issuing new credentials,
-  2. Validating existing credentials against a reference identifier, and
+  2. Validating a credential against a reference identifier, and
   3. Validating whether or not two credentials represent the same user.
 - A **Delivery Service**, described fully in {{delivery-service}}, provides
   methods for:
@@ -786,8 +792,7 @@ deployments for them to interoperate:
   (client) owned by the same application user. However, one device may control
   many signature keys but should still only be considered a single client.
 - A policy on how long to allow a member to stay in a group without updating its
-  leaf keys, and whether there should be any attempt to resolve a stale leaf key
-  before removing the member.
+  leaf keys before removing them.
 
 Finally, there are some additional application-defined behaviors that are
 partially an individual application's decision but may overlap with
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -338,7 +338,7 @@ with a large amount of trust and the compromise of one of its
 functionalities could allow an adversary to, among other things,
 impersonate group members. We discuss security considerations
 regarding the compromise of the different AS functionalities in detail
-in Section {{as-compromise}}.
+in {{as-compromise}}.
 
 The association between members' identities and signature keys is fairly
 flexible in MLS.  As noted above, there is no requirement that all clients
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-architecture.md
```diff
@@ -379,7 +379,7 @@ view of the security analysis.
 Upon joining the system, each client stores its initial cryptographic
 key material with the Delivery Service. This key material, called a
 KeyPackage, advertises the functional abilities of the client such as
-supported protocol versions and extensions and the following
+supported protocol versions, extensions, and the following
```
bifurcation, line 382:
```suggestion
supported protocol versions, supported extensions, and the following
```

draft-ietf-mls-architecture.md
```diff
@@ -696,6 +696,109 @@ software is updated while the client is a member of a group. Thus, in addition
 to preventing downgrade attacks, the members of a group can also observe when it
 is safe to upgrade to a new ciphersuite or protocol version.
 
+# Interoperability Requirements
```
bifurcation, line 699: It seems like these requirements apply not just for cases where two deployments need to interop, but even within a single deployment, in the sense that these are all choices that the deployment has to make.  Maybe something like:

```
# Operational Requirements

MLS is a security layer that needs to be integrated with an application. An
fully-functional deployment of MLS will have to make a number of decisions about
how MLS is configured and operated.  Deployments that wish to interoperate will
need to make compatible decisions. This section lists all of the dependencies of
an MLS deployment that are external to the protocol specification, but would
still need to be aligned within a given MLS deployment, or for two deployments
to potentially interoperate. 
```

bifurcation, line 699: It seems like these requirements apply not just for cases where two deployments need to interop, but even within a single deployment, in the sense that these are all choices that the deployment has to make.  Maybe something like:

```
# Operational Requirements

MLS is a security layer that needs to be integrated with an application. An
fully-functional deployment of MLS will have to make a number of decisions about
how MLS is configured and operated.  Deployments that wish to interoperate will
need to make compatible decisions. This section lists all of the dependencies of
an MLS deployment that are external to the protocol specification, but would
still need to be aligned within a given MLS deployment, or for two deployments
to potentially interoperate. 
```

draft-ietf-mls-architecture.md
```diff
@@ -696,6 +696,109 @@ software is updated while the client is a member of a group. Thus, in addition
...
+
+A fully functional deployment of MLS requires a number of infrastructure
+services to be available, as well as several policies and design trade-offs that
+are decided by the application. This section lists all of the dependencies of an
+MLS deployment that are external to the protocol specification, but would still
+need to be aligned for two deployments to potentially interoperate.
+
+MLS has the built-in ability to negotiate protocol versions, ciphersuites,
+extensions, and credential types. For two deployments to interoperate, they must
```
bifurcation, line 708: `Capabilities` also has proposal types.

bifurcation, line 708: `Capabilities` also has proposal types.

draft-ietf-mls-architecture.md
```diff
@@ -696,6 +696,109 @@ software is updated while the client is a member of a group. Thus, in addition
...
+have overlapping support in each of these categories.
+
+MLS relies on the following network services. These network services would
+either need to be the same, or compatible at the wire-format level, in order for
```
bifurcation, line 712:
```suggestion
need to be compatible in order for
```

bifurcation, line 712:
```suggestion
need to be compatible in order for
```

draft-ietf-mls-architecture.md
```diff
@@ -696,6 +696,109 @@ software is updated while the client is a member of a group. Thus, in addition
...
+two different deployments based on them to interoperate.
+
+- An **Authentication Service**, described fully in {{authentication-service}},
+  defines the types of credentials which may be used in a deployment and
+  provides methods for:
+  1. Issuing new credentials,
+  2. Validating existing credentials against a reference identifier, and
```
bifurcation, line 719:
```suggestion
  2. Validating a credential against a reference identifier, and
```
```suggestion
  2. Validating existing credentials against a reference identifier, and
```

bifurcation, line 719:
```suggestion
  2. Validating a credential against a reference identifier, and
```

draft-ietf-mls-architecture.md
```diff
@@ -696,6 +696,109 @@ software is updated while the client is a member of a group. Thus, in addition
...
+     for a user's own clients.
+- Additional services may or may not be required depending on the application
+  design:
+  - If assisted joining is desired (meaning that the ratchet tree is not
+    provided in Welcome messages), there must be a method to download the
+    ratchet tree corresponding to a group.
+  - If assisted joining is desired and the Delivery Service is not able to
+    compute the ratchet tree itself (because some proposals or commits are sent
+    encrypted), there must be a method for group members to publish the updated
+    ratchet tree after each commit.
```
bifurcation, line 735: You could summarize this in one bullet with something like "The tree can be either computed by the DS or uploaded by a group member after each Commit."

bifurcation, line 735: You could summarize this in one bullet with something like "The tree can be either computed by the DS or uploaded by a group member after each Commit."

draft-ietf-mls-architecture.md
```diff
@@ -696,6 +696,109 @@ software is updated while the client is a member of a group. Thus, in addition
...
+MLS requires the following parameters to be defined, which must be the same for
+two implementations to interoperate:
+
+- The maximum total lifetime that is acceptable for a KeyPackage.
+- How long to store the resumption secret for past epochs of a group.
+- The degree of tolerance that's allowed for out-of-order message delivery:
+  - How long to keep unused nonce and key pairs for a sender
+  - A maximum number of unused key pairs to keep.
+  - A maximum number of steps that clients will move a secret tree ratchet
+    forward in response to a single message before rejecting it.
```
bifurcation, line 759: These don't have to be the same, do they?  For example, if the DS doesn't allow out-of-order delivery, then you would never detect a difference.  Maybe just soften the "must be the same", e.g., to "must be compatible"

bifurcation, line 759: These don't have to be the same, do they?  For example, if the DS doesn't allow out-of-order delivery, then you would never detect a difference.  Maybe just soften the "must be the same", e.g., to "must be compatible"

draft-ietf-mls-architecture.md
```diff
@@ -696,6 +696,109 @@ software is updated while the client is a member of a group. Thus, in addition
...
+  - When proposals from external senders are allowed.
+  - When external joiners are allowed.
+- A policy for when two credentials represent the same client. Note that many
+  credentials may be issued authenticating the same identity but for different
+  signature keys, because each credential corresponds to a different device
+  (client) owned by the same application user. However, one device may control
+  many signature keys but should still only be considered a single client.
+- A policy on how long to allow a member to stay in a group without updating its
+  leaf keys, and whether there should be any attempt to resolve a stale leaf key
+  before removing the member.
```
bifurcation, line 790: What do you mean "resolve a stale leaf key"?  Like prompt the client to update?  I would just remove this last clause, or amend to something like "... and what actions should be taken with regard to stale leaf keys".

bifurcation, line 790: What do you mean "resolve a stale leaf key"?  Like prompt the client to update?  I would just remove this last clause, or amend to something like "... and what actions should be taken with regard to stale leaf keys".

draft-ietf-mls-architecture.md
bifurcation: It might be worth mentioning that `required_capabilities` can help ensure that things fail cleanly when required things aren't supported.

bifurcation (commented):
Really like this direction, few small tweaks.

draft-ietf-mls-architecture.md
beurdouche: I'll let you submit a subsequent PR if you think an additional action is needed.

beurdouche (approved):
This looks good.
Thanks Brendan, Thanks Richard!

PR #109, opened 2022-06-09
Note that consistent device sets can help security

### LINKED ISSUE #107: Add a section about cross-group identities
Particularly in multi-device scenarios (where a user has more than one device), we should recommend that clients correlate the set of devices of each user across groups. This would make it much harder for a malicious AS to issue fake device credentials for a particular user because clients would expect the credential to appear in all groups of which the user is a member. If a device credential does not appear in all groups after a while, clients have an indication that the credential might have been created without the user's knowledge.

This should however only be a recommendation since it might not be a desirable property in all scenarios. In practical terms, users might also need some time to add their new device to all of the groups, therefore a grace period might be needed. Taking the latter into account, correlating users' devices across groups is more of a detection mechanism than a prevention mechanism.

bifurcation:
Fixes #107

### DIFF
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -350,6 +350,21 @@ points in time, providing unlinkability and post-compromise security benefits.
 Some security trade-offs related to this flexibility are discussed in the
 security considerations.
 
+In many applications, there are multiple MLS clients that represent a single
+user, for example a human user with a mobile and desktop version of an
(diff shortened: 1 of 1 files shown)


raphaelrobert (approved):

kkohbrok (changes requested):
Made a quick pass. I think we should stick with "clients" rather than "devices".

beurdouche (approved):

PR #112, opened 2022-06-16
Review section 2.

Bren2010:


draft-ietf-mls-architecture.md
```diff
@@ -205,18 +205,11 @@ client components; and they may even involve some action by users.  For example:
...
-allows MLS clients to generate, redistribute and validate their
+allows MLS clients to generate, distribute, and validate
 credentials themselves.
-
-Similarly to the AS, the Delivery Service can be completely abstract
+Similar to the AS, the Delivery Service can be completely abstract
 if users are able to distribute credentials and messages without
-relying on a central Delivery Service. Note, though, that the MLS
-protocol requires group operation messages to be processed in-order by
-all MLS clients.
```
Bren2010, line 215: Removed because strict in-order processing isn't required

draft-ietf-mls-architecture.md
```diff
@@ -205,18 +205,11 @@ client components; and they may even involve some action by users.  For example:
...
-
-In some sense, a set of MLS clients which can achieve the AS and DS
-functionalities without relying on an external party do not need a
-Service Provider.
```
Bren2010, line 219: Service Provider is still needed, just a protocol instead of a literal company/person

draft-ietf-mls-architecture.md
```diff
@@ -234,12 +227,6 @@ Service Provider.
                            ..................................
 ~~~
 
-In many systems, the AS and the DS are actually operated by the same
-entity and may even be the same server. However, they are logically
-distinct and, in other systems, may be operated by different entities.
-Other partitions are also possible, such as having a separate
-directory functionality or service.
```
Bren2010, line 241: Removed because we just said the same thing above the picture

draft-ietf-mls-architecture.md
beurdouche: Is that true? They don't need to be received in order but they do need to be processed in the order of their epoch no? CC @bifurcation

draft-ietf-mls-architecture.md
```diff
@@ -312,16 +292,8 @@ In some messaging systems, clients belonging to the same user must
 all share the same signature key pair, but MLS does not assume this.
 
 Users will often use multiple devices, e.g., a phone as well as a laptop.
-Different devices may be represented as different clients, with independent
-cryptographic state, or they may share cryptographic state, relying on some
-application-provided mechanism to sync across devices.
```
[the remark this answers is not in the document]
beurdouche: @bifurcation Any opinion on this? I agree with Brendan that it is preferable not to share state but it is still a possibility...

draft-ietf-mls-architecture.md
```diff
@@ -312,16 +292,8 @@ In some messaging systems, clients belonging to the same user must
...
-
-The formal definition of a Group in MLS is the set of clients that
-have knowledge of the shared group secret established in the group key
-establishment phase of the protocol and have contributed to it.
-Until a Member has been added to the group and contributed to the group
-secret in a manner verifiable by other members of the group, other
-members cannot assume that the Member is a member of the group.
```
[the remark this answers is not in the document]
beurdouche: I remember the WG specifically discussing this and agreeing on this phrasing, so we should keep it.
Could you remove that specific change from the PR? (actually I can do it...)

draft-ietf-mls-architecture.md
```diff
@@ -312,16 +292,8 @@ In some messaging systems, clients belonging to the same user must
...
+Different devices are represented as different clients with independent
+cryptographic state.
```
beurdouche, line 296:
```suggestion
The formal definition of a Group in MLS is the set of clients that
have knowledge of the shared group secret established in the group key
establishment phase of the protocol and have contributed to it.
Until a Member has been added to the group and contributed to the group
secret in a manner verifiable by other members of the group, other
members cannot assume that the Member is a member of the group.
Different devices are represented as different clients with independent
cryptographic state.
```

beurdouche (approved):
Thanks Brendan! I'll merge this modulo the deletion of the paragraph regarding "being a member" since that was specifically discussed during an interim call.

draft-ietf-mls-architecture.md
Bren2010: This is discussed more in #113 but the state of different clients can fork based on the consistency model of the DS.

bifurcation: Seems like we can just generalize here, something like:

> Note, though, that in such scenarios, clients will need to implement logic that assures the delivery properties required of the DS (see {{delivery-guarantees}})

PR #113, opened 2022-06-16
Review sections 3-4.

Bren2010:
This PR pulls in the CAP theorem-related discussion from https://github.com/mlswg/mls-protocol/pull/697 which is a more helpful way to frame DS design.

PR #114, opened 2022-07-29
Use strongly/eventually consistent terminology.

Bren2010:


Bren2010:
Hey @rohan-wire, all the changes you requested should be here now.

draft-ietf-mls-architecture.md
```diff
@@ -316,7 +316,7 @@ The Authentication Service (AS) has to provide three functionalities:
    is valid with respect to a reference identifier
 
 3. Enable a group member to verify that a credential represents the same
```
beurdouche, line 318: I just realized that this is too restrictive. This can't be a requirement because some infras explicitly want to achieve the opposite. So this must be taken out of the "has to" and rephrased as... "Additionally, an AS can provide..."

draft-ietf-mls-architecture.md
```diff
@@ -402,7 +402,8 @@ security analysis.
 ## Key Storage and Retrieval
 
 Upon joining the system, each client stores its initial cryptographic
-key material with the Delivery Service. This key material, called a
+key material with the Delivery Service. Clients then continue adding new (and
+removing old) initial keying material on a regular basis. This key material, called a
```
beurdouche, line 406:
```suggestion
removing keying material on a regular basis. This key material, called a
```
(it is not necessarily old)

draft-ietf-mls-architecture.md
```diff
@@ -402,7 +402,8 @@ security analysis.
 ## Key Storage and Retrieval
 
 Upon joining the system, each client stores its initial cryptographic
-key material with the Delivery Service. This key material, called a
+key material with the Delivery Service. Clients then continue adding new (and
```
beurdouche, line 405:
```suggestion
key material with the Delivery Service. Clients then continue adding and
```

draft-ietf-mls-architecture.md
```diff
@@ -440,23 +441,29 @@ latter delivery pattern via unicast channels (sometimes known as "client
 fanout"), broadcast channels ("server fanout"), or a mix of both.
 
 For the most part, MLS does not require the Delivery Service to deliver messages
-in any particular order. The one exception is that, because an MLS group has a
-linear history, the members of the group must agree on the order in which
+in any particular order. Applications can set policies that control their
+tolerance for out-of-order messages (see {{operational-requirements}}), and
+messages that arrive significantly out-of-order will be dropped without
```
beurdouche, line 446:
```suggestion
messages that arrive significantly out-of-order can be dropped without
```

draft-ietf-mls-architecture.md
```diff
@@ -440,23 +441,29 @@ latter delivery pattern via unicast channels (sometimes known as "client
...
+otherwise affecting the protocol. There are two exceptions to this. First,
+Proposal messages must all arrive before the Commit that references them.
```
beurdouche, line 448:
```suggestion
Proposal messages should all arrive before the Commit that references them.
```

draft-ietf-mls-architecture.md
```diff
@@ -470,32 +477,38 @@ to detect. For instance, a Delivery Service can simply refuse to relay messages
...
+wait to apply it until it's broadcast back to them by the Delivery Service,
+assuming they don't receive another Commit first.
 
-The counter approach would be done to apply ordering to all messages, while the
-`epoch` approach would be done to only ensure an order on Commits.
+The Delivery Service can rely on the `epoch` and `content_type` fields of an
+MLSMessage to provide a guaranteed order only for handshake messages, and possibly
+even filter or reject redundant Commit messages proactively to prevent them from
+being broadcast. Alternatively, the Delivery Service could simply apply a
+guaranteed order to all messages and rely on clients to ignore redundant
```
beurdouche, line 497: I couldn't figure out what "guaranteed" means, could you rephrase please?

draft-ietf-mls-architecture.md
```diff
@@ -516,6 +529,10 @@ the rest of the group long enough that the mechanisms above no longer work. This
 can only be solved by sending a ReInit proposal to both groups, possibly with an
 external sender type, and recreating the group to contain all members again.
 
+If the Commit references an unknown proposal, group members may need to solicit
+the Delivery Service or other group members inidividually for the contents of
+the proposal.
```
beurdouche, line 534: to make sense of the proposal ?

beurdouche (changes requested):
LGTM otherwise.

draft-ietf-mls-architecture.md
rohan-wire: > I just realized that this is too restrictive. This can't be a requirement because some infras explicitly want to achieve the opposite. So this must be taken out of the "has to" and rephrased as... "Additionally, an AS can provide..."

I don't see the problem here. The requirement is that if two credentials represent the same instance, then other members can determine that. It does not require a system to have multiple instances. In fact if a system only has one instance then by this logic, the other members need to be able to determine that the one instance matches itself.

draft-ietf-mls-architecture.md
```diff
@@ -440,23 +441,29 @@ latter delivery pattern via unicast channels (sometimes known as "client
...
 
 For the most part, MLS does not require the Delivery Service to deliver messages
-in any particular order. The one exception is that, because an MLS group has a
-linear history, the members of the group must agree on the order in which
+in any particular order. Applications can set policies that control their
+tolerance for out-of-order messages (see {{operational-requirements}}), and
+messages that arrive significantly out-of-order will be dropped without
+otherwise affecting the protocol. There are two exceptions to this. First,
+Proposal messages must all arrive before the Commit that references them.
+Second, because an MLS group has a linear history of epochs, the members of the
```
rohan-wire, line 449: How about the requirement that a Commit creating an epoch arrives before any other messages which use that epoch?
[If that is not a strict requirement then we need to discuss how long a client will "hold" an application message for a newer epoch. There are potential DoS issues holding it too long.]

draft-ietf-mls-architecture.md
```diff
@@ -440,23 +441,29 @@ latter delivery pattern via unicast channels (sometimes known as "client
...
+group must agree on the order in which
 changes are applied.  Concretely, the group must agree on a single MLS Commit
 message that ends each epoch and begins the next one.
 
 In practice, there's a realistic risk of two members generating Commit messages
 at the same time, based on the same counter, and both attempting to send them to
 the group at the same time. The extent to which this is a problem, and the
 appropriate solution, depends on the design of the Delivery Service. Per the CAP
-theorem, there are two general classes of distributed system:
+theorem, there are two general classes of distributed system that the Delivery
```
rohan-wire, line 458: Minor:
- please add reference for CAP theorem
- a distributed DS in this context could be the clients and a single server, multiple (ex: federated servers), or clients and a network of servers. Probably worth mentioning that somewhere.

rohan-wire (commented):
Thanks Brendan, these changes address the most substantive comments I had about #113.

We also discussed bringing back some of sections 
`## Membership knowledge` - bringing back the section but just mentioning that anonymous MLS requires new work which could be addressed in future extensions.
and
`## Membership and offline members` - bring back the section please. I was going to propose text here about UpdatePath, frequency of updating keys, and removing "zombie" clients.

draft-ietf-mls-architecture.md
beurdouche: The notion of "instance" doesn't exist in MLS, so anyway this needs to change. Also it seems application specific so it should be removed entirely. Reading between the lines, I think you might mean something like:

"Additionally, in cases where applications require to identify relationships between credentials of the same user, the AS has to provide this functionality". Is this correct ?

Note, however, that it is not what is currently written. The current sentence makes it general to all credential, which we specifically don't want in MLS as some Applications might explicitly want to avoid that.

Bren2010: The concern this item was addressing was making sure it's possible to implement the requirements in [Section 13.2](https://messaginglayersecurity.rocks/mls-protocol/draft-ietf-mls-protocol.html#name-proposal-list-validation) of mls-protocol where a proposal list is invalid if:

```
* It contains multiple Add proposals that contain KeyPackages that
  represent the same client according to the application
* It contains an Add proposal with a KeyPackage that represents a
  client already in the group according to the application
```

How is the application supposed to implement these checks if there is not an Authentication Service API for it?

draft-ietf-mls-architecture.md
Bren2010: To get the proposal message, in the event that it was dropped or hasn't arrived yet

draft-ietf-mls-architecture.md
Bren2010: Added a note in Operational Requirements that messages may need to be buffered

draft-ietf-mls-architecture.md
beurdouche: I think the phrasing in the protocol doc might have to be changed because these are unnecessary restrictions on the protocol. As long as the key packages are different it is fine to add and remove the same user. It should be the responsibility of the application to signal this restriction through an extension. @bifurcation I think I can see what the intent was, is it to avoid issues on remove when double joining with different key package ?

Bren2010:
```
## Membership and offline members - bring back the section please. I was going to propose text here about UpdatePath, frequency of updating keys, and removing "zombie" clients.
```
You should definitely still add your new text, but consider doing it in the Operational Requirements or Security Considerations sections

draft-ietf-mls-architecture.md
beurdouche: @raphaelrobert @kkohbrok interested in your opinion too. My concern is about mandating the AS to say which keys are related.

rohan-wire: > Additionally, in cases where applications require to identify relationships between credentials of the same user, the AS has to provide this functionality". Is this correct ?
No. Not user. Instance of user.  

> I think the phrasing in the protocol doc might have to be changed because these are unnecessary restrictions on the protocol. As long as the key packages are different it is fine to add and remove the same user. It should be the responsibility of the application to signal this restriction through an extension. @bifurcation I think I can see what the intent was, is it to avoid issues on remove when double joining with different key package ?
No.

Benjamin, most IM systems support multiple clients/devices/instances for the same (often human) user. These are different keys, different identifiers, and they are treated by the application protocol, and by DoubleRatchet, differently. Applying MLS to these systems, each client/instance has it's own credential. Other MLS clients need to be able to tell when one client/instance wants to replace itself, and need to detect when the same client/instance tries to join twice (ex: because of a programming error, because it crashed at the wrong moment and lost state, or because it got very unlucky missing certain combinations of welcome or commit messages).

I someone wants to use the same credential on multiple devices running MLS, my first reaction would be that this is a very bad idea from a security perspective and out of scope of the MLS architecture document, but they are free to write an Internet draft describing their design to explain when/how/why it would be safe in their environment.

draft-ietf-mls-architecture.md
```diff
@@ -812,6 +816,10 @@ two implementations to interoperate:
   - A maximum number of unused key pairs to keep.
   - A maximum number of steps that clients will move a secret tree ratchet
     forward in response to a single message before rejecting it.
+  - Whether to buffer messages that aren't able to be understood yet due to
+    other messages not arriving first, and if so, how many to buffer. For
```
rohan-wire, line 820: propose:
s/how many to buffer/how many to buffer and for how long/

rohan-wire (commented):
Brendan, thanks for these modifications.

Any comment on the reintroducing the two removed subsections (Membership knowledge, and Membership and offline members)?

draft-ietf-mls-architecture.md
kkohbrok: I'm not entirely sure what are are discussing here: Do we want the AS to be able to tell

* if two individual clients (via their KeyPackages) belong to the same user, or 
* if two individual KeyPackages belong to the same client? 

I understand the quote from Section 13.2 to require the latter, although the change to the doc says something else.

My suggestion would thus be
```
3. Enable a group member to verify that a credential represents the same client as another credential
```
which would enable performing the checks mentioned by @Bren2010 and also, if I understand them correctly, address the concerns brought up by @rohan-wire.

beurdouche: From reading the discussion and the protocol doc again, I think this phrasing is fine, thank you. I think it might have been the intent but the phrasing was confusing. @Bren2010 Does that work for you?

raphaelrobert: Look like we are converging, but fwiw I think the AS should indeed be the component that links devices to accounts. For clients, it is important to track users/clients across groups as noted in #107 and introduced in #109. Note that I don't see any implication regarding privacy, because a well-designed AS would not leak information beyond what clients need to know.

Bren2010:
> Any comment on the reintroducing the two removed subsections (Membership knowledge, and Membership and offline members)?

You can bring the old text back with your new text if you'd like, but consider doing it in the Operational Requirements or Security Considerations sections because it doesn't go under Delivery Service

draft-ietf-mls-architecture.md
Bren2010: Wording changed to @kkohbrok's

draft-ietf-mls-architecture.md
```diff
@@ -77,26 +77,13 @@ informative:
        -
           ins: Google
 
-  Loopix:
```
rohan-wire, line 80: I think the goal was to add a new reference for the CAP theorem.  The reference for Loopix should remain.

rohan-wire (changes requested):
Please restore the Loopix reference and accept my propose wording for line 820.

Bren2010:
The loopix reference was removed because it's not used anywhere

rohan-wire:
@Bren2010 @beurdouche 
I think this PR is ready to merge. Are you waiting for anything at this point?

PR #116, opened 2022-08-21
WGLC comments to Functional Requirements section

rohan-wire:
WGLC comments to Functional Requirements Section:

- Include text for case when external joins are authorized by members autonomously
- Fix formatting of one RECOMMENDATION section
- Reference the mls-federation document (informational)

### DIFF
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -545,9 +545,11 @@ group.  When members perform changes directly, this is clearly the case.
 External joins are authorized indirectly, in the sense that a member publishing
 a GroupInfo object authorizes anyone to join who has access to the GroupInfo
 object.  External joins do not allow for more granular authorization checks to
-be done before the new member is added to the group, so if an application wishes
-to both allow external joins and enforce such checks, then the application will
-need to do such checks when a member joins, and remove them if checks fail.
+be done before the new member is added to the group, so if an application
+wishes to both allow external joins and enforce such checks, then either all
+the members of the group must all have the ability to check and reject invalid
+External joins autonomously, or the application needs to do such checks when a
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-architecture.md
```diff
@@ -545,9 +545,11 @@ group.  When members perform changes directly, this is clearly the case.
...
 a GroupInfo object authorizes anyone to join who has access to the GroupInfo
 object.  External joins do not allow for more granular authorization checks to
-be done before the new member is added to the group, so if an application wishes
-to both allow external joins and enforce such checks, then the application will
-need to do such checks when a member joins, and remove them if checks fail.
+be done before the new member is added to the group, so if an application
+wishes to both allow external joins and enforce such checks, then either all
+the members of the group must all have the ability to check and reject invalid
+External joins autonomously, or the application needs to do such checks when a
+member joins and remove them if those checks fail.
```
Bren2010, line 552: This seems to be saying the same thing twice? Did you mean: members need to be able to reject invalid external joins themselves, or the Delivery Service needs to be able to reject invalid external joins before they're sent to the group?

rohan-wire: either the client needs to reject invalid external joins, or the client will allow invalid joins and the application (unclear if it is part of the client or part of the DS) will remove invalidly joined clients. 

Note that originally, just the second option was present in the text. I edited the paragraph to provide the (in my opinion superior) option of the client simply rejecting an invalid external join.

bifurcation: I think this whole paragraph can be re-stated a little more clearly.  Filed #162 to handle this in follow-up.

bifurcation (approved):

PR #117, opened 2022-08-21
WGLC additions to Operational Requirements

rohan-wire:
WGLC comments on Operation Requirements section.

- Added several overlooked requirements

### DIFF
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -799,7 +799,7 @@ two implementations to interoperate:
     forward in response to a single message before rejecting it.
   - Whether to buffer messages that aren't able to be understood yet due to
     other messages not arriving first, and if so, how many and for how long. For
-    example, Commoit messages that arrive before a proposal they reference, or
+    example, Commit messages that arrive before a proposal they reference, or
     application messages that arrive before the Commit starting an epoch.
 
 MLS provides the following locations where an application may store arbitrary
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -752,16 +752,17 @@ interoperate.
 - An **Authentication Service**, described fully in {{authentication-service}},
   defines the types of credentials which may be used in a deployment and
   provides methods for:
-  1. Issuing new credentials,
-  2. Validating a credential against a reference identifier, and
-  3. Validating whether or not two credentials represent the same client.
-
+  1. Issuing new credentials with a relevant credential lifetime,
+  2. Validating a credential against a reference identifier,
+  3. Validating whether or not two credentials represent the same client, and
+  4. Optionally revoking credentials which are no longer authorized.
 - A **Delivery Service**, described fully in {{delivery-service}}, provides
   methods for:
   1. Delivering messages sent to a group to all members in the group.
   2. Delivering Welcome messages to new members of a group.
-  3. Downloading KeyPackages for specific clients, and uploading new KeyPackages
-     for a user's own clients.
+  3. Uploading new KeyPackages for a user's own clients.
+  4. Downloading KeyPackages for specific clients. Typically, KeyPackages are
+     used once and consumed.
 
 - Additional services may or may not be required depending on the application
   design:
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-architecture.md
```diff
@@ -815,14 +816,22 @@ MLS requires the following policies to be defined, which restrict the set of
 acceptable behavior in a group. These policies must be consistent between
 deployments for them to interoperate:
 
+- A policy on which ciphersuites are acceptable.
+- A policy on any mandatory or forbidden MLS extensions.
```
Bren2010, line 820: This isn't necessary since we negotiate ciphersuites and extensions, see line 740. Keep in mind these are policies that need to align for interop. We need some overlapped support for ciphersuites and extensions but will negotiate the rest, so not an interop issue

draft-ietf-mls-architecture.md
```diff
@@ -815,14 +816,22 @@ MLS requires the following policies to be defined, which restrict the set of
...
   limited to:
   - When a member is allowed to add or remove other members of the group.
   - When, and under what circumstances, a reinitialization proposal is allowed.
-  - When proposals from external senders are allowed.
-  - When external joiners are allowed.
+  - When proposals from external senders are allowed and how to authorize
+    those proposals.
+  - When external joiners are allowed and how to authorize those external
+    commits.
+  - Which other proposal types are allowed.
```
Bren2010, line 831: This is negotiated (line 740 again)

draft-ietf-mls-architecture.md
```diff
@@ -815,14 +816,22 @@ MLS requires the following policies to be defined, which restrict the set of
...
+- A policy of when members should commit pending proposals in a group.
```
Bren2010, line 832: The protocol doc specifies this to some extent (ie, you must commit proposals before you send an application message). To the extent that this behavior is not specified in the protocol doc, I don't believe further policy is necessary for interop

draft-ietf-mls-architecture.md
```diff
@@ -815,14 +816,22 @@ MLS requires the following policies to be defined, which restrict the set of
...
+- A policy of how to protect and share the GroupInfo objects needed for
+  external joins.
```
Bren2010, line 834: This is dup with line 776

draft-ietf-mls-architecture.md
rohan-wire: The DS in some contexts will have some requirements regarding ciphersuites and extensions.
For example, an organization which requires FIPS compliance cannot permit the use of any Edwards curves-based ciphersuites. Another might require a security level which allow p384, p521 and Ed448, but not Ed25519 or p256.
Likewise, a DS could forbid the use of the AppAck mechanism and require the content negotiation mechanism.
These are reasonable policy choices for legitimate operational or security reasons.

draft-ietf-mls-architecture.md
rohan-wire: This requires a policy. The protocol only says what clients have to do to make the protocol work, not what is desirable operationally. 
If you have a group with 5000 members and one member tries to leave the group by sending a Remove Proposal for itself, which client or clients Commit(s) the proposal? Do they wait indefinitely for one of them to send an application message (and then delay the message by having to send a Commit first)? Do they all immediately sending a Commit effectively causing a DoS attack on the DS? Does one member or a small number of them try the Commit?  When? According to what algorithm?
This is a serious operational concern.

draft-ietf-mls-architecture.md
rohan-wire: see the example of forbidding AppAck for operational reasons above.

draft-ietf-mls-architecture.md
rohan-wire: I did not read line 776 to include the concerns I had in mind in 833-834. For example, in a federated environment perhaps non-transport level encryption of the GroupInfo from the DS to the client that wants to join is required.

Bren2010: I would just consider encryption or no encryption part of the mechanism in line 776

draft-ietf-mls-architecture.md
rohan-wire: I don't agree that clients doing different things will necessarily work fine here. A client that naively sends a Commit instantly whenever it sees a pending proposal in a large group will not be allowed in some MLS networks.

draft-ietf-mls-architecture.md
rohan-wire: Honestly I read this section of the document several times and never got that impression. A lot of people are going to miss something that subtle and implicit. Even if some people might interpret line 776 as including this, what harm is there in making it clear in line 883-884?

seanturner:
Discussed on 12/8 interim, and folks seemed fine with this. I.e., it's ready to merge.

PR #120, opened 2022-08-27
Addressing I-D Nits

seanturner:
According to I-D Nits:

    Checking nits according to https://www.ietf.org/id-info/checklist :
    ----------------------------------------------------------------------------

    ** The abstract seems to contain references ([I-D.ietf-mls-protocol]),
       which it shouldn't.  Please replace those with straight textual mentions
       of the documents in question.

    ** The document seems to lack a both a reference to RFC 2119 and the
       recommended RFC 2119 boilerplate, even if it appears to use RFC 2119
       keywords. 

       RFC 2119 keyword, line 557: '...or forked group states MUST be deleted...'

### DIFF
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -87,7 +87,7 @@ informative:
 
 --- abstract
 
-The Messaging Layer Security (MLS) protocol {{!I-D.ietf-mls-protocol}}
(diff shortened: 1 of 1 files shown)


beurdouche (approved):
Thanks @seanturner

beurdouche:
TY!

PR #124, opened 2022-10-25
Review from OPSDIR.

### LINKED ISSUE #122: editorial nits from OPS-DIR review
For the authors' consideration:

-----

Section 3

    A member with a valid credential authenticates its MLS messages by signing them with the 

s/its/it's/

 mechanisms for a user to monitor and control which keys are associated to their identity.

 s/to their/with their/

 6. Operational Requirements

    A extension can help maintain interoperability

s/A extension/An extension/

    Whether there should be a extension in groups.

s/a extension/an extension/

7.2.2 Forward and Post-Compromise Security

    MLS partially defend against this problem by active member including freshness, however not 

s/defend/defends/

s/member/members/

    Mandate a key updates from clients that are not otherwise sending messages 

s/a key/key/

7.3.1 Compromise of AEAD key material

    authentication is not affected in neither of these 

s/neither/either/

7.3.2 Compromise of the Group Secrets of a single group for one or more group epochs

    of the compromised party with no ability to an honest group to recover message secrecy. 

s/to an/for an/

7.3.3 Compromise by an active adversary with the ability to sign messages

    the attacker can perform all operations which are available to an legitimate client 

s/an legitmate/a legitimate/

7.3.4 Compromise of the authentication with access to a signature key

    Beware that in both oracle and private key access, an active adaptive attacker, can follow th

s/attacker, can/attacker can/

7.3.5 Security consideration in the context of a full state compromise

    They also are providing the strong authentication guarantees to other clients

s/providing the strong/providing strong/

    hence we consider that their protection by additional security mechanism should be a priority.

s/mechanism/mechanisms/

    Overall there is no way to detect or prevent these compromise, as discussed 

s/compromise/compromises/

7.4.2.1 Privacy of delivery and push notifications

    For secure messaging systems, push notification are often sent real-time 

s/notification/notifications/

7.4.3 Authentication Service Compromise

    MLS client (which is under the control of the user) have often the ability 

s/have often/often have/

Bren2010:
Resolves #122

### DIFF
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -293,7 +293,7 @@ The Authentication Service (AS) has to provide three functionalities:
 3. Enable a group member to verify that a credential represents the same client
    as another credential
 
-A member with a valid credential authenticates it's MLS messages by signing them
+A member with a valid credential authenticates its MLS messages by signing them
 with the private key corresponding to the public key bound by its credential.
 
 The AS is considered an abstract layer by the MLS specification and part of this
@@ -1257,7 +1257,7 @@ these keys are extremely frequently used and changed with each message received
 by a client.
 
 However, the signature private keys are mostly used by clients to send a
-message. They are also providing strong authentication guarantees to other
+message. They also provide strong authentication guarantees to other
 clients, hence we consider that their protection by additional security
 mechanisms should be a priority.
 
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -293,7 +293,7 @@ The Authentication Service (AS) has to provide three functionalities:
 3. Enable a group member to verify that a credential represents the same client
    as another credential
 
-A member with a valid credential authenticates its MLS messages by signing them
+A member with a valid credential authenticates it's MLS messages by signing them
 with the private key corresponding to the public key bound by its credential.
 
 The AS is considered an abstract layer by the MLS specification and part of this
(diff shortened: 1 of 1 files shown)


rohan-wire (approved):
These changes look good to me.

draft-ietf-mls-architecture.md
```diff
@@ -293,7 +293,7 @@ The Authentication Service (AS) has to provide three functionalities:
 3. Enable a group member to verify that a credential represents the same client
    as another credential
 
-A member with a valid credential authenticates its MLS messages by signing them
+A member with a valid credential authenticates it's MLS messages by signing them
```
bifurcation, line 296:
```suggestion
A member with a valid credential authenticates its MLS messages by signing them
```

[This is incorrect.](https://www.merriam-webster.com/words-at-play/when-to-use-its-vs-its)  The MLS messages belong to the member, so we want the possessive here, not the contraction.

draft-ietf-mls-architecture.md
```diff
@@ -1257,11 +1257,11 @@ these keys are extremely frequently used and changed with each message received
 by a client.
 
 However, the signature private keys are mostly used by clients to send a
-message. They also are providing the strong authentication guarantees to other
+message. They are also providing strong authentication guarantees to other
```
bifurcation, line 1260:
```suggestion
message. They also provide strong authentication guarantees to other
```

bifurcation (approved):
Approved with nits.

PR #163, opened 2022-12-15
Add a mention of IPsec as a secure channel

### LINKED ISSUE #139: WireGuard but not IPsec?
_Comment by @paulwouters_

It is a bit odd to mention a non-IETF VPN protool, but not the IETF VPN protocol. Especially since IPsec also has constant byte stream support via IPTFS that WireGuard does not have :P

bifurcation:
Fixes #139


rohan-wire:
I'll add references for QUIC, TLS, IPsec, WireGuard, and TOR.

PR #164, opened 2022-12-15
State admission control possibilities more accurately

### LINKED ISSUE #152: Post-hoc remove of external joiners
_Comment by @paulwouters_

~~~ 
   External joins do
   not allow for more granular authorization checks to be done before
   the new member is added to the group, so if an application wishes to
   both allow external joins and enforce such checks, then the
   application will need to do such checks when a member joins, and
   remove them if checks fail.
~~~
This seems rather risky? What could such a member do in the short time between getting added and getting kicked? Would they see message history? Group membership? How many devices a user has? When members were last online? Why support such a group type? Wouldn't it be better to not support this and have an automated "special member" being a bot that handles member addition?

### LINKED ISSUE #162: Note the multiple ways to apply authorization policy to external joins
Following on from discussion in #116 

Even once a GroupInfo has been published to a prospective new joiner, there are still a few opportunities to apply authorization policy to the join:

1. The DS can block the external Commit from reaching the group members (requires intelligent DS)
2. The clients in the group can all refuse to apply the external Commit (requires consistent policy)
3. Any client can remove the member after allowing them to join (might expose some group comms to the joiner)

bifurcation:
Fixes #162 
Fixes #152

### DIFF
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -667,13 +667,9 @@ a GroupInfo object authorizes anyone to join who has access to the GroupInfo
 object. Both types of joins are done via a Commit message, which could be
 blocked by the DS or rejected by clients if the join is not authorized.  The
 former approach requires that Commits be visible to the DS; the latter approach
-requires that clients all share a consistent policy. In the event that an
-unauthorized member is able to join (e.g., via an un-checked external join), MLS
-enables any member to remove them.  If the application has each client detect
-and remove unauthorized clients before sending messages within the group, then
-an unauthorized joiner will not have access to anything confidential to the
-authorized members of the group, though they may have sent something to the
-group in the window between joining and removal.
+requires that clients all share a consistent policy. In the unfortunate event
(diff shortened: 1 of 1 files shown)


rohan-wire:
I think the spirit of this change is good, but the last two sentences are confusing talking about the client trying to join and an existing member trying to prevent them from seeing anything confidential after joining. 

My initial reaction to this idea was always that post-hoc removing someone who should never have joined is a terrible architectural decision. Post-hoc removal either allows an undesirable member a window to see messages sent by some members, or it requires consistent policy across members. If there is consistent policy across members the members should simply treat the Commit from the undesirable client as invalid. 
Rather than further condone post-hoc removal (which doesn't require any normative change to the MLS protocol), why not just delete mentions of post-hoc removal?

bifurcation:
Compromise - Deleted the sentence that talks about the security implications of post-hoc removal, left the mention of it simply being possible.

rohan-wire (approved):

PR #166, opened 2022-12-15
Minor fixes for AD review, part 1

### LINKED ISSUE #132: Section 2, bullet list
_Comment by @paulwouters_

~~~ 
   Clients may wish to do the following:
   [bullet list]
~~~
It is a little unclear whether this is the entire list of operations supported, or a subset. Maybe make that more explicit?

### LINKED ISSUE #133: Section 2, use of client and member
_Comment by @paulwouters_

The use of client, member and someone/everyone is a little inconsistent, eg see:

~~~ 
        For instance, any member can add or remove another client in a group.
~~~
I assume a member is identical to a client, but is a member of the group. But then the above "clients may wish to do the following:" should not be used with "remove one or more members from an existing group;"

Perhaps just adding a line saying "a member is a client that is within a group" might be useful? Or say that client and member are used interchangably, but that a client could be no member of any group ?

I guess this is further explained in [Section 2.1](https://www.ietf.org/archive/id/draft-ietf-mls-architecture-09.html#section-2.1) afterwards. It might be better to bring this up a little earlier?

### LINKED ISSUE #134: Served to who?
_Comment by @paulwouters_

~~~ 
        The Delivery Service is responsible for ensuring that each KeyPackage
        is only served once
~~~
It feels I am missing some context here. Who or whom is this KeyPackage served too?

### LINKED ISSUE #135: Are different extension lists supported ?
_Comment by @paulwouters_

~~~ 
        Clients may also want to support multiple protocol versions and
        ciphersuites.
~~~
What about different extensions? One client might be able to display images, while the other cannot ? Is that not possible, or was it just not listed here?

### LINKED ISSUE #136: What is a request?
_Comment by @paulwouters_

~~~ 
        The Delivery
        Service should provide the minimum number of KeyPackages necessary to
        satisfy a request.
~~~
Similarly, I feel I am missing context here. Why is minimum important, why need more than one? What is the definition of request?

bifurcation:
Fixes #132 
Fixes #133 
Fixes #134 
Fixes #135 
Fixes #136


draft-ietf-mls-architecture.md
```diff
@@ -1507,9 +1511,8 @@ For secure messaging systems, push notifications are often sent real-time as it
 is not acceptable to create artificial delays for message retrieval.
 
 > **RECOMMENDATION:**
-> If real time notifications are not necessary and that specific steps must be
-> taken to improve privacy, one can delay notifications randomly across
-> recipient devices using a mixnet or other techniques.
+> If real time notifications are not necessary , one can delay notifications
```
rohan-wire, line 1514: s/not necessary , one/not necessary, one/

rohan-wire (approved):

draft-ietf-mls-architecture.md
bifurcation, line 1514:
```suggestion
> If real time notifications are not necessary, one can delay notifications
```

PR #167, opened 2022-12-15
Minor fixes for AD review, part 2

### LINKED ISSUE #138: lost state and PSK
_Comment by @paulwouters_

~~~ 
   Group members whose local MLS state is lost or corrupted can
   reinitialize their state by re-joining the group as a new member and
   removing the member representing their earlier state.  An application
   can require that a client performing such a reinitialization prove
   its prior membership with a PSK.
~~~
Wouldn't this PSK that is the result of a membership be part of the "state lost or corrupted" ? What is the exact kind of failure mode one would address with this? Also, how often would such a PSK get updated as it needs to be known to add group members (and doing this too often would I assume lead to long dormant group members from losing access?) If one could remove once's old self without the PSK, than everyone could delete any member?

### LINKED ISSUE #140: 7.1 reads odd
_Comment by @paulwouters_

It states any "secure channel" can be used, but it is kinda says any insecure channel can be used too.

It also feels a little odd to describe these properties of MLS in the Security Considerations section, and not like in the introduction.

### LINKED ISSUE #142: In scope or not?
_Comment by @paulwouters_

~~~ 
      *RECOMMENDATION:* If unidirectional transport is used for the
      secure transport channel, prefer using a protocol which provides
      Forward Error Correction.
~~~
Is this saying a protocol within MLS, eg some application protocol? Or is this saying in this case, do not use MLS but another protocol at the messaging layer?

### LINKED ISSUE #145: 7.3 AEAD reference
_Comment by @paulwouters_

~~~ 
      Recall that the MLS protocol provides chains of AEAD keys
~~~
This "recall" is misplaced. It is not mentioned before until this point.

### LINKED ISSUE #146: 7.3.2 PCS, but ...
_Comment by @paulwouters_

~~~ 
   If the adversary is passive, it is expected from the PCS properties
   of the MLS protocol that, as soon as an honest Commit message is sent
   by the compromised party, the next epochs will provide message
   secrecy.
~~~
It feels a bit odd to think of a compromised user sending an honest Commit message. If the user doesn't know they are compromised, they will still be compromised and this new Commit message won't help. If the user did know they were compromised and undid that, then sure this honest Commit message helps. I think perhaps "honest Commit" should be expanded and say something about de-compromised client sending a Commit message. My issue with "honest" is that compromised user with a passive attacker is still "honest".

Similarly confusing is:

~~~ 
       an honest update from the compromised client happens
~~~

### LINKED ISSUE #149: Section 7.4.3
_Comment by @paulwouters_

~~~ 
   Infrastructures that provide cryptographic material or credentials in
   place of the MLS client (which is under the control of the user) have
   often the ability to use the associated secrets to perform operations
   on behalf of the user, which is unacceptable in many situations.
   Other mechanisms can be used to prevent this issue, such as the
   service blessing cryptographic material used by an MLS client.
~~~
This seems a bit of gloating about the MLS protocol/client. Why not just state why the attack is impossible and require that MLS clients do this? I don't see other features being compared to other/older/worse solutions?

### LINKED ISSUE #155: Deletion schedule
_Comment by @paulwouters_

~~~ 
   Messages sent
   during the client's membership are also secure as long as the client
   has properly implemented the MLS deletion schedule.
~~~
Is this the equivalent of proprietary IM clients "auto delete" or "view once" type messaging? "MLS deletation schedule" has not been defined before it is mentioned here. What is this?

### LINKED ISSUE #157: User expectations
_Comment by @paulwouters_

~~~ 
   Often, expectation from users is that the infrastructure will not
   retain the ability to constantly map the user identity to signature
   public keys of the MLS protocol.
~~~
Again I am confused there are users with expectations of the MLS protocol, which is still in a draft form only? Is this mixing MLS and non-MLS?

bifurcation:
Fixes #138 
Fixes #140 
Fixes #142 
Fixes #145 
Fixes #146 
Fixes #149 
Fixes #155 
Fixes #157

### DIFF
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -1288,8 +1288,8 @@ encryption keys for any AEAD chains and can encrypt and decrypt all messages for
 the compromised epochs.
 
 If the adversary is passive, it is expected from the PCS properties of the MLS
-protocol that, as soon as an honest Commit message is sent by the compromised
-party, the next epochs will provide message secrecy.
+protocol that, as soon as the compromised party remediates the compromise and
+sends an honest Commit message, the next epochs will provide message secrecy.
 
 If the adversary is active, the adversary can follow the protocol and perform
 updates on behalf of the compromised party with no ability for an honest group to
@@ -1314,8 +1314,8 @@ the compromise scenario implies a significant impact on both the secrecy and
 authentication guarantees of the protocol, especially if the attacker also has
 access to the group secrets. In that case both secrecy and authentication are
 broken.
-The attacker can generate any message, for the current and future epochs until
-an honest update from the compromised client happens.
+The attacker can generate any message, for the current and future epochs, until
+the compromise is remediated and the formerly compromised client sends an honest update.
 
 Note that under this compromise scenario, the attacker can perform all
 operations which are available to a legitimate client even without access to
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -1104,7 +1104,7 @@ of service for MLS.
 
 > **RECOMMENDATION:**
 > If unidirectional transport is used for the secure transport channel, prefer
-> using a protocol which provides Forward Error Correction.
+> using a transport protocol which provides Forward Error Correction.
 
 ## Intended Security Guarantees
 
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -1658,13 +1658,11 @@ confidentiality guarantees, it is a serious issue for privacy.
 > stored encrypted at rest and during the execution. Applications should also
 > consider anonymous systems for server fanout (for example {{Loopix}}).
 
-Often, expectation from users is that the infrastructure will not retain the
-ability to constantly map the user identity to signature public keys of the MLS
-protocol. Some infrastructures will keep a mapping between signature public keys
-of clients and user identities. This can benefit an adversary that has
-compromised the AS (or required access according to regulation) the ability of
-monitoring unencrypted traffic and correlating the messages exchanged within the
-same group.
+Some infrastructures will keep a
+mapping between signature public keys of clients and user identities. This can
+benefit an adversary that has compromised the AS (or required access according
+to regulation) the ability of monitoring unencrypted traffic and correlating the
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-architecture.md
```diff
@@ -1650,13 +1658,11 @@ confidentiality guarantees, it is a serious issue for privacy.
...
-protocol. Some infrastructures will keep a mapping between signature public keys
-of clients and user identities. This can benefit an adversary that has
-compromised the AS (or required access according to regulation) the ability of
-monitoring unencrypted traffic and correlating the messages exchanged within the
-same group.
+Some infrastructures will keep a
+mapping between signature public keys of clients and user identities. This can
+benefit an adversary that has compromised the AS (or required access according
+to regulation) the ability of monitoring unencrypted traffic and correlating the
+messages exchanged within the same group.
```
rohan-wire, line 1665: how about:
Some infrastructure keeps a mapping between keys used in the MLS protocol and user identities. An attacker with access to this information due to regulation or compromise can associated unencrypted group messages (ex: Commits and Proposals) with the corresponding user identity.

rohan-wire (changes requested):

draft-ietf-mls-architecture.md
```diff
@@ -1536,12 +1541,15 @@ incorrect or attacker-provided identities to clients.
 
 - The attacker can publish or distribute credentials
 
-Infrastructures that provide cryptographic material or credentials in place of
-the MLS client (which is under the control of the user) often have the ability
-to use the associated secrets to perform operations on behalf of the user, which
-is unacceptable in many situations. Other mechanisms can be used to prevent this
-issue, such as the service blessing cryptographic material used by an MLS
-client.
+In the past, some systems have had a centralized server geenrate signature key
```
bifurcation, line 1544:
```suggestion
In the past, some systems have had a centralized server generate signature key
```

draft-ietf-mls-architecture.md
bifurcation, line 1665:
```suggestion
Some infrastructure keeps a mapping between keys used in the MLS protocol and
user identities. An attacker with access to this information due to compromise
or regulation can associate unencrypted group messages (e.g., Commits and
Proposals) with the corresponding user identity.
```

PR #169, opened 2022-12-15
Discuss the privacy of a user's device set

### LINKED ISSUE #126: Each client is its own entity
_Comment by @paulwouters_

A user with multiple devices becomes a member of the group multiple times. This implies that group members can detect which of the user's clients was used for a specific message. Can this property be eliminated or at least reduced in the possible cases (eg if the user's devices are all online, could these work together to hide the information from which device/client the user sent the message? eg by always picking one of them to relay?)

bifurcation:
Fixes #126

### DIFF
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -1195,6 +1195,28 @@ may be possible to operate MLS in ways that provide certain deniability
 properties, but defining the specific requirements and resulting notions of
 deniability requires further analysis.
 
+### Associating a User's Clients
+
+When the same user uses multiple clients, it may be possible for other members
+of a group to recognize all of those clients as belonging to the same user.  For
+example, all of a user's clients might present credentials authenticating the
+user's identity.  This association among devices might be considered a leak of
+private information.  An application that wishes to mitigate this privacy risk
+has a couple of options.
+
+This risk only arises when the leaf nodes for the clients in question provide
+data that can be used to correlate the clients.  So one way to mitigate this
+risk is by only doing device-level authentication within MLS. If user-level
+authentication is still desirable, the application would have to be provide it
+through some other mechanism.
+
+It is also possible to maintain user-level authentication while hiding
+information about the clients that a user uses.  This can be done by having the
+clients share cryptographic state, so that they appear as a single client within
+the MLS group.  The application would need to provide a synchronization
+mechanism so that the clients' state remained consistent across changes to the
+MLS group.
+
 ## Endpoint Compromise
 
 The MLS protocol adopts a threat model which includes multiple forms of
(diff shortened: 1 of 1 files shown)


rohan-wire (approved):

draft-ietf-mls-architecture.md
```diff
@@ -1195,6 +1195,28 @@ may be possible to operate MLS in ways that provide certain deniability
...
 deniability requires further analysis.
 
+### Associating a User's Clients
+
+When the same user uses multiple clients, it may be possible for other members
+of a group to recognize all of those clients as belonging to the same user.  For
+example, all of a user's clients might present credentials authenticating the
+user's identity.  This association among devices might be considered a leak of
+private information.  An application that wishes to mitigate this privacy risk
+has a couple of options.
```
ekr, line 1205:
```suggestion
private information.  The remainder of this section describes several approaches for addressing this.
```

draft-ietf-mls-architecture.md
```diff
@@ -1195,6 +1195,28 @@ may be possible to operate MLS in ways that provide certain deniability
...
+
+This risk only arises when the leaf nodes for the clients in question provide
+data that can be used to correlate the clients.  So one way to mitigate this
+risk is by only doing device-level authentication within MLS. If user-level
```
ekr, line 1209: Is this text trying to say that the credentials only contain device identity?

ekr (changes requested):

draft-ietf-mls-architecture.md
beurdouche, line 1209:
```suggestion
risk is by only doing client-level authentication within MLS. If user-level
```

draft-ietf-mls-architecture.md
```diff
@@ -1195,6 +1195,28 @@ may be possible to operate MLS in ways that provide certain deniability
...
+authentication is still desirable, the application would have to be provide it
+through some other mechanism.
+
+It is also possible to maintain user-level authentication while hiding
+information about the clients that a user uses.  This can be done by having the
```
beurdouche, line 1214:
```suggestion
information about the clients that a user owns.  This can be done by having the
```

draft-ietf-mls-architecture.md
```diff
@@ -1195,6 +1195,28 @@ may be possible to operate MLS in ways that provide certain deniability
...
+clients share cryptographic state, so that they appear as a single client within
+the MLS group.  The application would need to provide a synchronization
+mechanism so that the clients' state remained consistent across changes to the
+MLS group.
+
```
beurdouche, line 1219:
```suggestion

>**RECOMMENDATION:**
> Avoid sharing cryptographic state between clients to improve resilience against compromises.
> An attacker could use one compromised device to establish ownership of a state across
> other devices and reduce the ability of the user to recover.
```

beurdouche (approved):

draft-ietf-mls-architecture.md
beurdouche: This should be per client not per device.

PR #171, opened 2022-12-16
Reshuffle sentences in Ghost user discussion for better flow

rohan-wire:
The Section `#### Authentication compromise: Ghost users and impersonations` has three recommendations, but the discussion rambles. The proposed text places topics related to each recommendation before that recommendation.

### DIFF
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -1564,19 +1564,13 @@ emitted credentials might be compromised.
 #### Authentication compromise: Ghost users and impersonations
 
 One important feature of MLS is that all
-clients know who is in the group at all times. This means that - if all Members
-of the group and the Authentication Service are honest - no other parties than
+Members know which other clients are in the group at all times. If all Members
+of the group and the Authentication Service are honest, no parties other than
 the members of the current group can read and write messages protected by the
-protocol for that Group. Note that when a `basic` Credential is used, the
-Authentication Service also needs an out-of-band mechanism to verify the
-identity asserted in the Credential.
+protocol for that Group.
 
-Beware though, the link between the cryptographic identity of the Client and the
-real identity of the User is important. With some Authentication Service
-designs, a private or centralized authority can be trusted to generate or
-validate signature keypairs used in the MLS protocol. This is typically the
-case in some of the biggest messaging infrastructures.
-
+Beware though, the link between the cryptographic
+identity of the client and the real identity of the End User is important.
 Details about how to verify the identity of a client depend on the MLS
 Credential type used. For example, cryptographic verification of credentials
 can be largely performed autonomously on the clients for the `x509`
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-architecture.md
```diff
@@ -1564,36 +1564,37 @@ emitted credentials might be compromised.
...
-
-Beware though, the link between the cryptographic identity of the Client and the
-real identity of the User is important. With some Authentication Service
-designs, a private or centralized authority can be trusted to generate or
-validate signature keypairs used in the MLS protocol. This is typically the
-case in some of the biggest messaging infrastructures.
+protocol for that Group.
 
+Beware though, the link between the cryptographic
+identity of the client and the real identity of the End User is important.
```
ekr, line 1573: I would actually just remove this sentence, which doesn't seem helpful.

draft-ietf-mls-architecture.md
```diff
@@ -1564,36 +1564,37 @@ emitted credentials might be compromised.
...
-the expected one. Because a user can have many MLS clients running the MLS
-protocol, it possibly has many signature keypairs for multiple devices.
-
 > **RECOMMENDATION:**
 > Select the strongest MLS Credential type available among the
 > target members of an MLS group.
 
+While the Authentication Service is often very well protected from external
+attackers, it might be the case that this service is compromised. In such
+infrastructure, the AS could validate a (or generate a new) signature keypair for
```
ekr, line 1587:
```suggestion
If the AS is compromised, it could validate a (or generate a new) signature keypair for
```

draft-ietf-mls-architecture.md
```diff
@@ -1564,36 +1564,37 @@ emitted credentials might be compromised.
...
+an identity which is not the expected one. Because a user can have many MLS
```
ekr, line 1588:
```suggestion
an attacker. Because a user can have many MLS
```

draft-ietf-mls-architecture.md
```diff
@@ -1564,36 +1564,37 @@ emitted credentials might be compromised.
...
+clients running the MLS protocol, it possibly has many signature keypairs
+for multiple devices. Note that when a `basic` Credential is used, the
```
ekr, line 1590: Is the point that this might be difficult to detect?

draft-ietf-mls-architecture.md
```diff
@@ -1564,36 +1564,37 @@ emitted credentials might be compromised.
...
+Authentication Service also needs an out-of-band mechanism to verify the
+identity asserted in the Credential.
+
+With some Authentication Service designs, a private or centralized authority
+is trusted to validate signature keypairs used in the MLS protocol. This is
+typically the case in some of the biggest messaging infrastructures.
```
ekr, line 1596: This paragraph just comes out of nowhere. What's the point?

ekr (changes requested):

draft-ietf-mls-architecture.md
kkohbrok: I agree that it's not clear what the reader should do with this information. At the same time this is at the heart of proper e2e authentication in person-to-person messaging. Maybe something that would better fit in the authentication service section?

draft-ietf-mls-architecture.md
```diff
@@ -1564,36 +1564,37 @@ emitted credentials might be compromised.
 #### Authentication compromise: Ghost users and impersonations
 
 One important feature of MLS is that all
-clients know who is in the group at all times. This means that - if all Members
-of the group and the Authentication Service are honest - no other parties than
+Members know which other clients are in the group at all times. If all Members
```
beurdouche, line 1567:
```suggestion
Members know which other members are in the group at all times. If all Members
```

draft-ietf-mls-architecture.md
beurdouche: This is already in the AS section so I'll remove the sentence.

beurdouche, line 1573:
```suggestion
```

draft-ietf-mls-architecture.md
beurdouche, line 1590:
```suggestion
for multiple devices. These attacks could be very difficult to detect.

> **RECOMMENDATION:**
> Provide a key transparency mechanism for the Authentication Services to allow public
> verification of the credentials authenticated by this service.

Note that when a `basic` Credential is used, the
```

draft-ietf-mls-architecture.md
```diff
@@ -1564,36 +1564,37 @@ emitted credentials might be compromised.
...
+an identity which is not the expected one. Because a user can have many MLS
+clients running the MLS protocol, it possibly has many signature keypairs
+for multiple devices. Note that when a `basic` Credential is used, the
+Authentication Service also needs an out-of-band mechanism to verify the
+identity asserted in the Credential.
+
+With some Authentication Service designs, a private or centralized authority
+is trusted to validate signature keypairs used in the MLS protocol. This is
+typically the case in some of the biggest messaging infrastructures.
+
```
beurdouche, line 1597:
```suggestion

```

draft-ietf-mls-architecture.md
beurdouche: Doesn't look like it but I added a sentence for that particular remark.

beurdouche (approved):

beurdouche:
Thanks!

PR #182, opened 2023-02-01
add a Terminology subsection and glossary

rohan-wire:
Several comments from IESG review suggested including important terms early in the document.
This PR adds a short glossary.

### DIFF
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -284,7 +284,7 @@ A _Commit_ message initiates a new epoch by instructing members of the group to
 implement a collection of proposals. Proposals and Commits are collectively
 called _Handshake messages_.
 A _KeyPackage_ provides keys that can be used to add the client to a group,
-including its LeafNode, and _Signature Key_.
+including an initiation LeafNode, and _Signature Key_.
 A _Welcome_ message provides a new member to the group with the information to
 initialize their state for the epoch in which they were added.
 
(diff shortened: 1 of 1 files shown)


ekr:
I absolutely hate having glossaries upfront, so I'd prefer not to do this. If we think there are terms that are not clear, let's define them in context.

rohan-wire:
> I absolutely hate having glossaries upfront, so I'd prefer not to do this. If we think there are terms that are not clear, let's define them in context.

I checked in a revision in a narrative style. How is that?

draft-ietf-mls-architecture.md
```diff
@@ -261,12 +261,44 @@ all users, for all group sizes, even when it reduces to only two users.
...
+identity, credentials, and capabilities.
+
+Various messages are used in the evolution from epoch to epoch.
+A _Proposal_ message proposes
+a change to be made in the next epoch, such as adding or removing a member.
+A _Commit_ message initiates a new epoch by instructing members of the group to
+implement a collection of proposals. Proposals and Commits are collectively
+called _Handshake messages_.
+A _KeyPackage_ provides keys that can be used to add the client to a group,
+including its LeafNode, and _Signature Key_.
```
br-hale, line 287: Perhaps consider replacing "including its LeafNode" with "including an initiation LeafNode"? It seems a little ambiguous if just "LeafNode", i.e. how that relates to use in other trees.

rohan-wire, line 287:
```suggestion
including an initiation LeafNode, and _Signature Key_.
```

rohan-wire:
Replaced by #188 because of merge conflicts on this branch.

Issue #179, opened 2023-01-31
Review from Éric Vyncke (INT)


beurdouche:
```
Éric Vyncke has entered the following ballot position for
draft-ietf-mls-architecture-10: Discuss

When responding, please keep the subject line intact and reply to all
email addresses included in the To and CC lines. (Feel free to cut this
introductory paragraph, however.)


Please refer to https://www.ietf.org/about/groups/iesg/statements/handling-ballot-positions/ 
for more information about how to handle DISCUSS and COMMENT positions.


The document, along with other ballot positions, can be found here:
https://datatracker.ietf.org/doc/draft-ietf-mls-architecture/



----------------------------------------------------------------------
DISCUSS:
----------------------------------------------------------------------


# Éric Vyncke, INT AD, comments for draft-ietf-mls-architecture-10
CC @evyncke

Thank you for the work put into this document.

I find it very difficult to understand as many concepts are not explained.
Usually, I start reading the architecture I-D before the protocol one, and this
document does not help understanding the architecture. We could even argue
whether the I-D is about architecture or about security considerations.

Please find below some blocking DISCUSS points (easy to address), some
non-blocking COMMENT points (but replies would be appreciated even if only for
my own education), and some nits.

Thanks to Tatuya Jinmei, the Internet directorate reviewer (at my request),
please consider this int-dir review with one nit worth considering:
https://datatracker.ietf.org/doc/review-ietf-mls-architecture-10-intdir-telechat-jinmei-2023-01-29/
and as noted I share Jinmei's view about the clarity of the I-D.

Please note that Dave Lawrence is the DNS directorate reviewer (at my request)
and you may want to consider this dnsdir review as well when Dave will complete
the review (no need to wait for it though):
https://datatracker.ietf.org/doc/draft-ietf-mls-architecture/reviewrequest/16998/

Special thanks to Sean Turner for the shepherd's detailed write-up including
the WG consensus *and* the justification of the intended status.

I hope that this review helps to improve the document,

Regards,

-éric

## DISCUSS

As noted in https://www.ietf.org/blog/handling-iesg-ballot-positions/, a
DISCUSS ballot is a request to have a discussion on the following topics:

### Section 6

What is a `LeafNode` ? I.e., please define what it is before using the term ;-)
Very similar issue in 5.1 for `GroupInfo`. And as noted by Jinmei "Commit" and
"Proposal".

### Section 7.1

Is it appropriate for an IETF RFC (even if informal) to qualify WireGuard and
[...]

Issue #178, opened 2023-01-31
Review from Roman Danyliw (SEC)


beurdouche:
```
Roman Danyliw has entered the following ballot position for
draft-ietf-mls-architecture-10: Discuss

When responding, please keep the subject line intact and reply to all
email addresses included in the To and CC lines. (Feel free to cut this
introductory paragraph, however.)

Please refer to https://www.ietf.org/about/groups/iesg/statements/handling-ballot-positions/ 
for more information about how to handle DISCUSS and COMMENT positions.

The document, along with other ballot positions, can be found here:
https://datatracker.ietf.org/doc/draft-ietf-mls-architecture/

----------------------------------------------------------------------
DISCUSS:
----------------------------------------------------------------------
```

- [ ]
```
** Section 7.3.5
     *RECOMMENDATION:* If the threat model of the system is against an
     adversary which can access the messages on the device without even
     needing to attack MLS, the application should delete plaintext
     messages and ciphertexts immediately after encryption or
     decryption.

How does this recommend work relative to making the system usable? 
Specifically, how does the application fulfill its functional purpose if the
plaintext should be deleted after encryption or decryption.  When does the user
get a chance to read the message?  Or when does the autonomous client get a
chance to parse the plaintext and take appropriate action on it?
```

- [ ]
```
** Section 7.4.3.2.  Combined, these two recommendations are providing an
inconsistent recommendation:

(a) Section 7.4.3.1

     *RECOMMENDATION:* Always use encrypted group operation messages to
     limit privacy risks.

(b) Section 7.1.2.
     *RECOMMENDATION:* Never use the unencrypted mode for group
     operations without using a secure channel for the transport layer.

Recommendation-(a) is saying “always used encrypted group operation messages”,
but Recommendation-(b) is allowing for unencrypted ones as long as transport
security is used.
```

- [ ]
```
** Section 7.4.3.1

*RECOMMENDATION:* Select the strongest MLS Credential type
     available among the target members of an MLS group.

What is the metric for “strongest”? Is there an absolute the rank order of
“strongest” to “weakest”?  Is this decision application specific?
```

```
----------------------------------------------------------------------
COMMENT:
----------------------------------------------------------------------

Thank to Yoav Nir for the SECDIR review.
```
- [ ]
```
** (Similar feedback as provided by Yoav Nir and Éric Vyncke) Based on just the
title of this document, I assumed that it would introduce me to the general
concepts and design of MLS.  It surprised me that the narrative style of text
[...]

Issue #176, opened 2023-01-16
Review from Yoav Nir


beurdouche:
Reviewer: Yoav Nir
Review result: Has Nits

I have previously done an early review of version -09 of this draft.

At the time I was impressed by the breadth and depth of the security
considerations but found a few nits.

In version -10, the breadth and depth are still there with that section taking
no fewer than 18 pages in the text version. Unfortunately, the nits are also
still present.

rohan-wire:
All addressed in PR #183 

NITS from review of version -9
> The document describes the Message Layer Security (MLS) architecture, and
> includes a very thorough and comprehensive analysis of the security and privacy
> issues in section 7, including recommendations. Considering the length and the
> breadth, it is not surprising that I cannot find anything missing.
> 
> A few nits:
> 
> The document uses terms that are not defined within the document:
> * "MLSCiphertext" appears in section 7.1.1 without having been defined earlier.
> That's probably acceptable because it is defined in the protocol document which
> is a normative reference. Still, such terms are commonly redefined. *
> "Anonymous credentials" are mentioned without having been defined in either
> this document or in the protocol document. It *is* a term that is used
> occasionally, but AFAICT the only IETF document that defines it is [RFC 4949](https://datatracker.ietf.org/doc/rfc4949/),
> which is (a) not referenced here, and (b) defines it as a USG term, and (c)
> says that Internet drafts "SHOULD NOT use this term"
> 
> Section 7.2.3 defines "deniability" and then says that "MLS does not make any
> claims with regard to deniability". So why do we need that paragraph?
> 
> The document includes three instances of the adverb "extremely" and five of the
> adverb "very", all in section 7, and I don't think they are necessary. Better
> yet, some of them could be replaced with actual measures. For example, "MLS
> avoids needing to send the full list of recipients ... because that list is
> potentially extremely large in MLS." How large? 5? 1000? A billion?  In other
> cases, "...clients have the extremely important role of deleting appropriate
> keys..." just sounds like a new formulation of SHOULD. I don't think those
> adverbs are necessary. Interestingly, the protocol document has only one
> instance of "very" and it is expanded to a measure: "The use of variable-size
> integers for vector lengths allows vectors to grow very large, up to 2^30
> bytes."

beurdouche:
Verified as resolved. Thank you !

PR #183, opened 2023-02-01
Address comments and NITS from SEC and INT area

rohan-wire:
Address IESG comments from Roman Danyliw and Yoav Nir (SEC)
Also address IESG telechat comments from Éric Vyncke (INT) because they overlap substantially

Partially addresses #178 (Roman Danyliw) - all DISCUSS issues except one already addressed in PR #180 (merged)
addresses some comments from #179 (Éric Vyncke) - DISCUSS issues both addressed 
Completely addresses #176 (Yoav Nir)
Addresses DISCUSS in #184 (Zaheduzzaman Sarker)

### DIFF
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -1090,10 +1090,10 @@ metadata if it cannot compromise the secure channel.
 ### Integrity and Authentication of Custom Metadata
 
 The MLS protocol provides an authenticated "Additional Authenticated Data" field
-for applications to make data available outside the MLSCiphertext.
+for applications to make data available outside a PrivateMessage.
 
 > **RECOMMENDATION:** Use the "Additional Authenticated Data" field of the
-> MLSCiphertext message instead of using other unauthenticated means of sending
+> PrivateMessage instead of using other unauthenticated means of sending
 > metadata throughout the infrastructure. If the data is private, the
 > infrastructure should use encrypted Application messages instead.
 
@@ -1120,10 +1120,10 @@ connections. Such a system helps in preventing anonymous clients from sending
 arbitrary numbers of group operation messages to the Delivery Service or the MLS
 clients.
 
-> **RECOMMENDATION:** Anonymous credentials can be used in order to help DoS
+> **RECOMMENDATION:** Use credentials uncorellated to specific users to help DoS
 > attacks prevention, in a privacy preserving manner. Note that the privacy of
 > these mechanisms has to be adjusted in accordance with the privacy expected
-> from the secure transport links. (See more discussion further down.)
+> from secure transport links. (See more discussion in the next section.)
 
 ### Message Suppression and Error Correction
 
@@ -1244,10 +1244,10 @@ to a third party. This is often called "non-repudiation".
 
 Roughly speaking, "deniability" is the opposite of "non-repudiation", i.e., the
 property that it is impossible to prove to a third party that a message was sent
-by a given sender.  MLS does not make any claims with regard to deniability.  It
-may be possible to operate MLS in ways that provide certain deniability
-properties, but defining the specific requirements and resulting notions of
-deniability requires further analysis.
+by a given sender. Deniability is not currently provided by the MLS protocol, but the
+protocol avoids constraints that would make it impossible to add deniability
+properties (via extensions) in the future. Defining the specific requirements
+and resulting notions of deniability requires further analysis.
 
 ### Associating a User's Clients
 
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -578,7 +578,7 @@ However, note that a malicious Delivery Service could also reorder messages or
 provide an inconsistent view to different users.  The "generation" counter in
 MLS messages provides per-sender loss detection and ordering that cannot be
 manipulated by the DS.  A mechanism for more robust protections is discussed in
-{{?I-D.ietf-mls-extensions}}.  A DS can cause a partition in the group by
+{{Section 2.1 of I-D.ietf-mls-extensions}}.  A DS can cause a partition in the group by
 partitioning key exchange messages; this can be detected only by out of band
 comparison (e.g., confirming that all clients have the same
 `epoch_authenticator` value`).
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-architecture.md
```diff
@@ -1608,8 +1616,8 @@ incorrect or attacker-provided identities to clients.
 
 - The attacker can publish or distribute credentials
 
-In the past, some systems have had a centralized server generate signature key
-pairs and distribute them to clients.  In such cases, the centralized server is
+It is possible for a system to have a centralized server generate signature key
```
Bren2010, line 1619:
```suggestion
It is possible for a system to have a central server generate signature key
```

draft-ietf-mls-architecture.md
```diff
@@ -1641,7 +1649,8 @@ read and write messages protected by the protocol for that Group.
 
 Details about how to verify the identity of a client depend on the MLS
 Credential type used. For example, cryptographic verification of credentials can
-be largely performed autonomously on the clients for the `x509` Credential
+be largely performed autonomously (e.g. without additional network traffic) by
```
Bren2010, line 1652:
```suggestion
be largely performed automatically by
```
Trying to convey that X.509 is not manual, not that it doesn't require network (even though it may not)

draft-ietf-mls-architecture.md
```diff
@@ -1243,10 +1250,10 @@ to a third party. This is often called "non-repudiation".
 
 Roughly speaking, "deniability" is the opposite of "non-repudiation", i.e., the
 property that it is impossible to prove to a third party that a message was sent
-by a given sender.  MLS does not make any claims with regard to deniability.  It
-may be possible to operate MLS in ways that provide certain deniability
-properties, but defining the specific requirements and resulting notions of
-deniability requires further analysis.
+by a given sender. Deniability is not currently provided by the MLS protocol, but the
+protocol avoids constraints that would make it impossible to add deniability
+properties (via extensions) in the future. Defining the specific requirements
```
br-hale, line 1255: Recommend striking "in the future". This seems to imply that the potential for deniability is linked to follow-on iterations of MLS and cannot be achieved in its current state (i.e. based on implementation decisions). Note that if kept, "in the future" does not imply only to the current time but to the full lifetime of the standard.

rohan-wire, line 1255:
```suggestion
properties (via extensions). Defining the specific requirements
```

draft-ietf-mls-architecture.md
rohan-wire: I was trying to convey that it doesn't require any additional network traffic, as this was the genesis of the original comment/question.

draft-ietf-mls-architecture.md
rohan-wire: Keeping centralized for consistency (appears twice more in the same paragraph)

draft-ietf-mls-architecture.md
Bren2010: X.509 verification does require network traffic though, to check revocation. The "In contrast" below is talking about out-of-band / manual verification

ekr:
On Wed, Feb 8, 2023 at 10:37 AM Brendan McMillion ***@***.***>
wrote:

> ***@***.**** commented on this pull request.
> ------------------------------
>
> In draft-ietf-mls-architecture.md
> <https://github.com/mlswg/mls-architecture/pull/183#discussion_r1100536500>
> :
>
> > @@ -1641,7 +1649,8 @@ read and write messages protected by the protocol for that Group.
>
>  Details about how to verify the identity of a client depend on the MLS
>  Credential type used. For example, cryptographic verification of credentials can
> -be largely performed autonomously on the clients for the `x509` Credential
> +be largely performed autonomously (e.g. without additional network traffic) by
>
> X.509 verification does require network traffic though, to check
> revocation. The "In contrast" below is talking about out-of-band / manual
> verification
>

This is not entirely correct. For instance, modern browsers increasingly
use centrally distributed revocation information, and so this does not
require network traffic to check revocation.

-Ekr

draft-ietf-mls-architecture.md
```diff
@@ -301,9 +305,10 @@ may even involve some action by users.  For example:
 * MLS clients connected to a peer-to-peer network could instantiate a
   decentralized DS by transmitting MLS messages over that network.
 
-* In an MLS group using a PKI for authentication, the AS would comprise the
-  certificate issuance and validation processes, both of which involve logic
-  inside MLS clients as well as various servers.
+* In an MLS group using a Public Key Infrastructure (PKI) for authentication,
+  the AS would comprise the certificate issuance and validation processes,
+  both of which involve logic inside MLS clients as well as various
+  existing PKI roles (ex: Certificate Authorities).
```
seanturner, line 311:
```suggestion
  existing PKI roles (ex: Certification Authorities).
```

bifurcation: Please change "ex:" to "e.g.,"

draft-ietf-mls-architecture.md
```diff
@@ -242,8 +245,9 @@ Instructions are on that page as well.  Editorial changes can be
 managed in GitHub, but any substantive change should be discussed on
 the MLS mailing list.
 
-End-to-end security is a requirement for instant messaging systems and is
-commonly deployed in many such systems. In this context, "end-to-end" captures
+End-to-end security is a used in the vast majority of instant messaging systems,
+and also deployed in systems for other purposes such as calling and conferencing.
+In this context, "end-to-end" captures
```
raphaelrobert, line 250: This is where we should reference https://datatracker.ietf.org/doc/draft-knodel-e2ee-definition/ instead of having our own definition

draft-ietf-mls-architecture.md
rohan-wire, line 1652:
```suggestion
be largely performed autonomously (e.g. without user interaction) by
```

draft-ietf-mls-architecture.md
rohan-wire: This paragraph is not a definition. It is a statement of adoption.  The only aspect of e2e we specifically call out is that the operator of the messaging system can be considered an attacker. 

As for including this reference anyway, looking at the status, this has several Not Ready telechat ballot positions, so I am inclined to leave it out.

PR #190, opened 2023-02-10
Add link to e2e security definition draft

kkohbrok:
As briefly discussed during the interim, this PR adds a link to [draft-knodel-e2ee-definition](https://datatracker.ietf.org/doc/draft-knodel-e2ee-definition/). I hope I got the markdown link formatting right.

### DIFF
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -251,7 +251,7 @@ In this context, "end-to-end" captures
 the notion that users of the system enjoy some level of security -- with the
 precise level depending on the system design -- even in the face of malicious
 actions by the operator of the messaging system. See
-{{!I-D.draft-knodel-e2ee-definition}} for a more detailed definition of
+{{?I-D.draft-knodel-e2ee-definition}} for a more detailed definition of
 end-to-end security.
 
 Messaging Layer Security (MLS) specifies an architecture (this document) and a
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -250,7 +250,9 @@ and also deployed in systems for other purposes such as calling and conferencing
 In this context, "end-to-end" captures
 the notion that users of the system enjoy some level of security -- with the
 precise level depending on the system design -- even in the face of malicious
-actions by the operator of the messaging system.
+actions by the operator of the messaging system. See
+{{!I-D.draft-knodel-e2ee-definition}} for a more detailed definition of
+end-to-end security.
 
 Messaging Layer Security (MLS) specifies an architecture (this document) and a
 protocol {{!I-D.ietf-mls-protocol}} for providing end-to-end security in this


ekr:
I was not in the interim, but I do not think that we should land this PR.

I've reviewed that document quite carefully and I don't think that the definition it provides is helpful in this context, in part because it seems to want to pull in stuff outside the protocol.

draft-ietf-mls-architecture.md
```diff
@@ -250,7 +250,9 @@ and also deployed in systems for other purposes such as calling and conferencing
 In this context, "end-to-end" captures
 the notion that users of the system enjoy some level of security -- with the
 precise level depending on the system design -- even in the face of malicious
-actions by the operator of the messaging system.
+actions by the operator of the messaging system. See
+{{!I-D.draft-knodel-e2ee-definition}} for a more detailed definition of
```
rohan-wire, line 254:
```suggestion
{{?I-D.draft-knodel-e2ee-definition}} for a more detailed definition of
```

rohan-wire:
> I was not in the interim, but I do not think that we should land this PR.
> 
> I've reviewed that document quite carefully and I don't think that the definition it provides is helpful in this context, in part because it seems to want to pull in stuff outside the protocol.

I also don't think this PR is a good idea. Defining in depth what end-to-end encryption is, is not especially relevant to the this document. The only salient feature of end-to-end encryption for MLS is that messaging is confidential to intermediaries who are part of the DS function.

raphaelrobert:
The discussion above is precisely why we should reference this, to keep the debate about what E2EE is out of the MLSWG. draft-knodel-e2ee-definition has seen a fair bit of discussion and is following the IETF process. The intention is simply to punt all discussion about E2EE to that document. Rohan mentioned that the previous version of the draft has "not ready" ballots, @mallory what is the current status of draft-08?

mallory:
Paul is reviewing the latest version that has addressed the last round of comments. It's changed significantly and especially the formal definition section is much shorter, which I think lends itself well as a definitional reference. Furthermore the rest of the document does invoke more user facing features of an MLS system for consideration.

ekr:
> The discussion above is precisely why we should reference this, to keep the debate about what E2EE is out of the MLSWG. draft-knodel-e2ee-definition has seen a fair bit of discussion and is following the IETF process. The intention is simply to punt all discussion about E2EE to that document.

This doesn't have that effect. It has the effect of this document *endorsing* that definition.


>  Rohan mentioned that the previous version of the draft has "not ready" ballots, @mallory what is the current status of draft-08?

To be clear:
There are no ballots because this has not been brought to the IESG or even had a LC. What there are are *Reviews*, a number of which are not positive (including mine).

As Rohan says, the only salient feature here is the technical one that the content is encrypted between the endpoints, and if we need to state any such definition, that is what we should do here.

mallory:
But that misremembers the discussion in MLS that the architecture draft would not try to define end to end encryption. 

The draft definition isn't done. If it doesn't meet folks' satisfaction there is an opportunity to change the draft.

ekr:
> But that misremembers the discussion in MLS that the architecture draft would not try to define end to end encryption.

I don't recall that discussion one way or the other and I suspect I was not part of it, so I don't think "misremember" is quite the right word. In any case, the draft does not need to define E2EE. It simply needs to state the technical properties that MLS provides, which it already does. IOW, we can simply close this PR and be done.


> The draft definition isn't done. If it doesn't meet folks' satisfaction there is an opportunity to change the draft.

Perhaps, but at the end of the day this citation has to go to a specific version and the -08 version isn't satisfactory.

rohan-wire:
Hi Mallory,
The way I read the relevant 2 paragraphs in the mls-architecture, we are
not trying to define end-to-end encryption, and I think that is
appropriate. What I think the document is trying to say is that we consider
threats/attacks by intermediaries.

I guess there are two questions here:
1) Should mls-architecture define or endorse a definition of e2e
encryption, or not?
In one of the interims, I'm pretty sure you are correct that we said we
would create our own definition. I don't recall consensus that we would
endorse a definition, but I remember it coming up.

2) Is the document consistent with that decision?
If you think there is language in mls-architecture that implies a
definition, I would be happy to learn more about your concerns
(chat/call/async) so we can fix that.

Many thanks,
-rohan

*Rohan Mahy  *l  Vice President Engineering, Architecture

Chat: @rohan_wire on Wire



Wire <https://wire.com/en/download/> - Secure team messaging.

*Zeta Project Germany GmbH  *l  Rosenthaler Straße 40,
<https://maps.google.com/?q=Rosenthaler+Stra%C3%9Fe+40,%C2%A0+10178+Berlin,%C2%A0+Germany&entry=gmail&source=g>10178
Berlin,
<https://maps.google.com/?q=Rosenthaler+Stra%C3%9Fe+40,%C2%A0+10178+Berlin,%C2%A0+Germany&entry=gmail&source=g>
Germany
<https://maps.google.com/?q=Rosenthaler+Stra%C3%9Fe+40,%C2%A0+10178+Berlin,%C2%A0+Germany&entry=gmail&source=g>

Geschäftsführer/Managing Director: Alan Duric

HRB 149847 beim Handelsregister Charlottenburg, Berlin

VAT-ID DE288748675

ekr:
I don't believe that the draft in question is going to be advancing as an IETF specification. At this point, I think we can close this PR without action.

PR #196, opened 2023-07-09
Update text around Clients and Groups for clarity

ekr:


draft-ietf-mls-architecture.md
```diff
@@ -388,31 +388,33 @@ controller who can modify the group. MLS is compatible with having group
...
+messages to and from a given user identically regardless of which
+client they are associated with, or may choose to distinguish them.
+
+When a client is part of a Group, it is called a Member.  A group in
+MLS is defined as the set of clients that have knowledge of the shared
+group secret established in the group key establishment phase of the
+protocol and have contributed to it.  Until a Member has been added to
+the group and contributed to the group secret in a manner verifiable
+by other members of the group, other members cannot assume that the
+Member is a member of the group.
```
ekr, line 417: @bifurcation I'm not sure how helpful this last sentence is (it is in the original). I suppose the idea here is that (for instance) the Welcome message could be corrupt or the new member has lost their key. However, it seems like as a practical matter, when Bob learns that Charlie has been added to the group, then it is reasonable for him to assume that he is.

After all, Charlie could lose his key *after* being added to the group. Thoughts?

beurdouche: I am gonna slightly change the sentence, but there were concerns in the past about not having this. The idea was that anyone can be added to a group without their content. So we want to consider clients to be members only when the Welcome information was processed and the client has made an update.

ekr: I think the problem here is that the phrase "member of the group" does not have an intuitive meaning

Like in this case, Charlie (1) is getting copies of the message and (2) can read them, so how is he not a member of the group

Isn't this just the standard two generals stuff.

beurdouche: Well the definition we used until now is that it is not a group member until it has contributed to the epoch secret.
This is not amazing, I agree, but this is the current status-quo... What we wanted to avoid is that if I take one of your KeyPackages and add you to "the evil group" you are automatically considered a member of the group.

PR #201, opened 2023-08-15
Remove essentially all uses of "tree".

ekr:
In Eric Vyncke's DISCUSS he correctly notes that LeafNode is used without being defined. I fixed that by just removing the term but more generally the problem is the use of "tree" without it being defined. Obviously the tree is an important concept but that's really a matter for mls-protocol not mls-architecture which is just concerned with group state as an abstract structure.

This change removes essentially all uses of the term tree except for:

- In the final security considerations section where I think it's OK in context.

- In this sentence

  "A maximum number of steps that clients will move a secret tree ratchet forward in response to a single message before rejecting it."

I'm pretty sure that this can be reworded but I wasn't quite sure the best way.

### DIFF
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -1019,13 +1019,16 @@ interoperate.
 
 - Additional services may or may not be required depending on the application
   design:
-  - If assisted joining is desired (meaning that the group state is not
-    provided in Welcome messages), there must be a method to download the
-    group state corresponding to a group.
-  - If assisted joining is desired and the Delivery Service is not able to
-    compute the group state itself (because some proposals or commits are sent
-    encrypted), there must be a method for group members to publish the updated
-    group state after each commit.
+
+   - Instead of distributing group information to a new members
+     entirely within a Welcome messages, an application may rely on
+     servers to store public information about the group, such as
+     group members' credentials and related public keys, and provide
+     this information to new members. (In such a system, Welcome
+     messages are still necessary to convey confidential information
+     about the group.) Such an application will require mechanisms for
+     new members to download the current information for a group, and
+     for members to update the information as the group evolves.
   - If external joiners are allowed, there must be a method to publish a
     serialized `GroupInfo` object (with an `external_pub` extension) that
     corresponds to a specific group and epoch, and keep that object in sync with
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -1068,7 +1068,7 @@ deployments to interoperate:
 
 - Group IDs, as decided by group creators and used to uniquely identify a group.
 
-- Associated with the per-user group state (specifically in
+- Application-level identifiers of public key material (specifically
   the `application_id` extension as defined in {{Section 5.3.3 of ?RFC9420}}).
 
 MLS requires the following policies to be defined, which restrict the set of
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -1019,13 +1019,13 @@ interoperate.
 
 - Additional services may or may not be required depending on the application
   design:
-  - If assisted joining is desired (meaning that the ratchet tree is not
+  - If assisted joining is desired (meaning that the group state is not
     provided in Welcome messages), there must be a method to download the
-    ratchet tree corresponding to a group.
+    group state corresponding to a group.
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-architecture.md
```diff
@@ -1019,13 +1019,13 @@ interoperate.
...
+  - If assisted joining is desired (meaning that the group state is not
     provided in Welcome messages), there must be a method to download the
-    ratchet tree corresponding to a group.
+    group state corresponding to a group.
   - If assisted joining is desired and the Delivery Service is not able to
-    compute the ratchet tree itself (because some proposals or commits are sent
+    compute the group state itself (because some proposals or commits are sent
     encrypted), there must be a method for group members to publish the updated
-    ratchet tree after each commit.
+    group state after each commit.
```
rohan-wire, line 1028: I think these cases all refer to the `ratchet_tree` extension, as opposed to the ratchet tree generally. I suggest leaving these as is.

draft-ietf-mls-architecture.md
```diff
@@ -1019,13 +1019,13 @@ interoperate.
 
 - Additional services may or may not be required depending on the application
   design:
-  - If assisted joining is desired (meaning that the ratchet tree is not
+  - If assisted joining is desired (meaning that the group state is not
```
rohan-wire, line 1022:
```suggestion
  - If assisted joining is desired (meaning that the `ratchet_tree` extension is not
```

draft-ietf-mls-architecture.md
```diff
@@ -1019,13 +1019,13 @@ interoperate.
...
     provided in Welcome messages), there must be a method to download the
-    ratchet tree corresponding to a group.
+    group state corresponding to a group.
```
rohan-wire, line 1024:
```suggestion
    ratchet tree corresponding to a group.
```

draft-ietf-mls-architecture.md
```diff
@@ -1019,13 +1019,13 @@ interoperate.
...
   - If assisted joining is desired and the Delivery Service is not able to
-    compute the ratchet tree itself (because some proposals or commits are sent
+    compute the group state itself (because some proposals or commits are sent
```
rohan-wire, line 1026:
```suggestion
    compute the group state itself, including the corresponding `ratchet_tree`
    object, (for example, because some proposals or commits are sent
```

rohan-wire (changes requested):

draft-ietf-mls-architecture.md
```diff
@@ -1068,7 +1068,8 @@ deployments to interoperate:
 
 - Group IDs, as decided by group creators and used to uniquely identify a group.
 
-- The `application_id` extension of a LeafNode.
+- Associated with the per-user group state (specifically in
```
Bren2010, line 1071:
```suggestion
- Application-level identifiers of public key material (specifically
```

bifurcation (commented):
I'm less offended by "ratchet tree" here and more by "assisted joining", which is not defined elsewhere.  So instead of the point edit @ekr has proposed here, I would suggest rewriting the two "assisted join" bullets entirely, something like:

> Instead of distributing group information to a new members entirely within a Welcome messages, an application may rely on servers to store public information about the group, such as group members' credentials and related public keys, and provide this information to new members.  (In such a system, Welcome messages are still necessary to convey confidential information about the group.)  Such an application will require mechanisms for new members to download the current information for a group, and for members to update the information as the group evolves.

ekr:
I adopted Richard's suggestion here. @rohan-wire @bifurcation PTAL.

rohan-wire:
> I'm less offended by "ratchet tree" here and more by "assisted joining", which is not defined elsewhere. So instead of the point edit @ekr has proposed here, I would suggest rewriting the two "assisted join" bullets entirely, something like:
> 
> > Instead of distributing group information to a new members entirely within a Welcome messages, an application may rely on servers to store public information about the group, such as group members' credentials and related public keys, and provide this information to new members.  (In such a system, Welcome messages are still necessary to convey confidential information about the group.)  Such an application will require mechanisms for new members to download the current information for a group, and for members to update the information as the group evolves.

Fair enough, but this is the section on Operational Considerations, which is essentially a checklist of policy knobs for people who implement and deploy MLS. By this point in the document, these people better know what an assisted join is or we are doing them a disservice.

beurdouche (approved):

PR #202, opened 2023-08-16
Add new Terminology section.

ekr:


draft-ietf-mls-architecture.md
```diff
@@ -272,13 +272,44 @@ all users, for all group sizes, including groups of only two clients.
 
 # General Setting
 
+## Terminology
```
ekr, line 275: Perhaps rename this to "Protocol Overview"?

draft-ietf-mls-architecture.md
```diff
@@ -272,13 +272,44 @@ all users, for all group sizes, including groups of only two clients.
...
-group. A user might have multiple MLS clients associated with the
-same identity, for instance if they have different devices.
+group. As groups change membership and group or member properties, they
+advance from one _epoch_ to another and the cryptographic state of the
+group evolves.
+
+The group is represented using a _ratchet tree_, which represents the members
+as the leaves of a tree. It is used to efficiently encrypt to subsets of the
+members. Each member has a _LeafNode_ object in the tree holding the client's
+identity, credentials, and capabilities.
```
ekr, line 289: @rohan-wire I think this goes back to our discussion on PR #201, so before I propose changes, let's see if we can agree on general principles and then we can make both this and PR#201 match.

My feeling is that we should:

1. Not force people to be aware of the notion that this is a tree, as this is confusing without context, but rather talk about the "group state"
2. Where appropriate reference the "ratchet_tree" value but as carrying the group state.

An approximate rubric, then, would be to use the term "ratchet_tree" but never "ratchet tree" (sans underscore). 

WDYT of that?

draft-ietf-mls-architecture.md
```diff
@@ -272,13 +272,44 @@ all users, for all group sizes, including groups of only two clients.
...
+
+Various messages are used in the evolution from epoch to epoch.
+A _Proposal_ message proposes
+a change to be made in the next epoch, such as adding or removing a member.
+A _Commit_ message initiates a new epoch by instructing members of the group to
+implement a collection of proposals. Proposals and Commits are collectively
+called _Handshake messages_.
+A _KeyPackage_ provides keys that can be used to add the client to a group,
+including its LeafNode, and _Signature Key_.
```
ekr, line 298: If you agree with me above, then we can remove "LeafNode" here and replace it with something like "encryption key, credentials, and signature key"

draft-ietf-mls-architecture.md
```diff
@@ -272,13 +272,44 @@ all users, for all group sizes, including groups of only two clients.
...
+A _Welcome_ message provides a new member to the group with the information to
+initialize their state for the epoch in which they were added.
+
+Of course most (but not all) applications use MLS to send encrypted group messages.
+An _application message_ is an MLS message with an arbitrary application payload.
+
+Finally, a _PublicMessage_ contains an integrity-protected MLS Handshake message,
+while a _PrivateMessage_ contains a confidential, integrity-protected Handshake
+or Application message.
```
ekr, line 307:
```suggestion
or application message.
```

You don't capitalize this above.

draft-ietf-mls-architecture.md
```diff
@@ -272,13 +272,44 @@ all users, for all group sizes, including groups of only two clients.
...
+
+For a more
+detailed explanation of these terms, please consult the MLS protocol specification.
```
ekr, line 310:
```suggestion
detailed explanation of these terms, please consult the MLS protocol specification {{RFC9420}}.
```

ekr (commented):
@rohan-wire per our discussion last night, I took a look at this. I suggested some changes and I think there's one cross-PR topic to discuss. I think once we resolve these we should be able to land this.

draft-ietf-mls-architecture.md
rohan-wire: Sure

draft-ietf-mls-architecture.md
rohan-wire: This depends on who the primary audience is for the architecture document. Is it primarily for implementers and architects (who need to understand the ratchet tree and its implications), or primarily for operators (who can largely ignore the ratchet tree as a detail). That said, even for operators, deciding whether Welcome and GroupInfo objects contain a `ratchet_tree` extension or it gets that separately (and how) is something the operator needs to understand.

I think the idea that MLS uses a type of tree to efficiently reach different subsets/subtrees is important for the first category.

Bren2010 (approved):

draft-ietf-mls-architecture.md
bifurcation: It seems to me that the important thing to convey here is that there's a data structure here that has credentials for members (plus maybe some related public keys).  Rather than invent some new terminology, I would be inclined to just call it a ratchet tree, and maybe say that the description of each member is in a leaf node, but not call out LeafNode specifically.

On another, minor note, the ratchet tree doesn't represent the whole state of the group, since it doesn't capture the key schedule / secret tree that the clients share. 

Net of all that, I would rephrase this para something like the following:

> The membership of the group is represented using a _ratchet tree_, in which each member is represented by a leaf node that describes the member's identity, credential, and capabilities.  The tree also contains public keys that allow for efficient updates to the group.

And if we do this, I would amend my comment on #201 to replace "public information about the group, such as group members' credentials and related public keys" with "ratchet tree".

ekr: I would suggest that we remove the term "ratchet" from tree here. Everyone knows what a tree is and ratchet isn't needed here. In the interest of progress I guess I can live with LeafNode.

ekr:
@bifurcation @rohan-wire if you are happy with this, I think it is ready to land.

rohan-wire (approved):

beurdouche (approved):

PR #204, opened 2023-09-27
Retrieve initial keying material from DS, not AS

thomdixon:
Section 5 denotes the DS is responsible for providing the initial keying material: 

> As a directory service providing the initial keying material for clients to use. This allows a client to establish a shared key and send encrypted messages to other clients even if they're offline.

### DIFF
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -404,7 +404,7 @@ sender can use a new key.
 
 ## Step 3: Adding Bob go the Group
 
-When Alice wants to create a group including Bob, she first uses the AS to look
+When Alice wants to create a group including Bob, she first uses the DS to look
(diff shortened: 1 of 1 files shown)


Bren2010 (approved):

draft-ietf-mls-architecture.md
```diff
@@ -404,7 +404,7 @@ sender can use a new key.
 
 ## Step 3: Adding Bob go the Group
```
beurdouche, line 405:
```suggestion
## Step 3: Adding Bob to the Group
```

beurdouche (approved):

beurdouche:
Thanks @thomdixon @Bren2010 !

thomdixon:
Thanks for the quick review and fix @beurdouche @Bren2010 🙏 (and all your work on the protocol, ofc)!

PR #218, opened 2023-11-07
Clarify that groupinfos are also subject to access control. Fixes #209

### LINKED ISSUE #209: Text around External Joins
> With both mechanisms, changes to the membership are initiated from inside the
group.  When members perform changes directly, this is clearly the case.
External joins are authorized indirectly, in the sense that a member publishing
a GroupInfo object authorizes anyone to join who has access to the GroupInfo
object. 

Presumably other group members could actually reject the commit when it happens, right? Perhaps we need to adjust this text slightly. @bifurcation

ekr:


beurdouche (approved):

beurdouche:
@ekr Feel free to merge.

rohan-wire (approved):

PR #224, opened 2023-11-07
Remove recommendation not to have the AS generated keys

ekr:
This removes the text suggesting that the AS not generate keys. I agree that this is the best practice, but it's just such a standard practice that i think even mentioning the alternative creates confusion.

### DIFF
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -1799,6 +1799,13 @@ emitted credentials might be compromised.
 > ability of an adversary with no physical access to extract the top-level
 > signature private key.
 
+Note that historically some systems generate signature keys on the
+Authentication Service and distribute the private keys to clients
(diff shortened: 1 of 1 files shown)


beurdouche (changes requested):
I don't think this should be removed. To my knowledge, there are enough non-PKI driven use cases to justify this graph. 

Maybe we can attenuate the language to make this "acceptable" ?

ekr:
In what use cases does the AS generate the key pairs?

rohan-wire:
Maybe we could just strengthen the text to say basically:

"Some service did this in the past. DON'T DO THIS!!!! (really!)"

ekr:
See above.

rohan-wire (approved):

beurdouche:
I misread Ekr's comment in my first review. We actually all agree.

PR #230, opened 2023-11-08
Aead key compromise cleanup

ekr:


ekr:
This builds on PR#229, so that should be reviewed first.

draft-ietf-mls-architecture.md
```diff
@@ -1089,7 +1089,7 @@ interoperate.
     is useful for clients that do not have the ability to send the full public
     state in a Welcome message when inviting auser or for client that need to
     recover from a loss of their state. Such public state can contain privacy
-    sensitive information such as group members' credentials and related public
+    sensitive information such as group members' credentrials and related public
```
rohan-wire, line 1092:
```suggestion
    sensitive information such as group members' credentials and related public
```

draft-ietf-mls-architecture.md
```diff
@@ -1406,25 +1405,6 @@ the scope of this document.
...
-property that it is impossible to prove to a third party that a message was sent
-by a given sender.  MLS does not make any claims with regard to deniability.  It
-may be possible to operate MLS in ways that provide certain deniability
-properties, but defining the specific requirements and resulting notions of
-deniability requires further analysis.
-
-### Associating a User's Clients
-
-When the same user uses multiple clients, it may be possible for other members
-of a group to recognize all of those clients as belonging to the same user.  For
```
rohan-wire, line 1427: This looks like a mistake.  was the intention to delete the entire Non-repudiation vs Deniability discussion?

And then the "Associating a User's Clients" section heading is removed and the second sentence of that section is now under the previous heading???

rohan-wire (changes requested):
Let's finish PR #229 first, then merge that update into this one. I don't think it is possible to review this PR yet because of the overlapping changes with the other PR.

draft-ietf-mls-architecture.md
ekr: Pilot error. This is fixed in #229, so when that PR is done i will rebase.

beurdouche (approved):

PR #232, opened 2023-11-09
Issue223 mixnets

ekr:


rohan-wire:
I suggest we leave the mixnet recommendation. It is not difficult to implement on a client and it the only practical solution possible for users who really require this type of privacy protection (typically against state actors).

PR #241, opened 2024-01-19
Move sig key reuse outside of the privacy section

beurdouche:


draft-ietf-mls-architecture.md
```diff
@@ -1908,11 +1911,7 @@ In certain cases, the adversary can access specific bindings between public keys
 and identities. If the signature keys are reused across groups, the adversary
 can get more information about the targeted user.
 
-> **RECOMMENDATION:** Do not use the same signature keypair across
-> groups. Update all keys for all groups on a regular basis. Do not preserve
-> keys in different groups when suspecting a compromise.
-
-> **RECOMMENDATION:** Separate the service binding the identities and the public
+> **RECOMMENDATION:** Separate the binding the identities and the public
```
rohanmahy, line 1914: s/Separate the binding the identities and the public/Separate the binding of the identities and the corresponding public/

rohan-wire:
> Again, do we really expect people not to use the same signature keypair across groups? Do people actually do that?

There are some clients that allow pseudonyms for different groups. This is mostly in large open-ish groups. I agree most clients don't do that in the general case.

PR #245, opened 2024-01-26
Remove reference to otherwise undefined "assisted" joining. Fixes #244

### LINKED ISSUE #244: What is assisted joining?
I genuinely don't remember what this term means and it is used in the architecture draft but not defined

ekr:


beurdouche:
Thanks !

PR #246, opened 2024-01-30
Explain the impact of unencrypted operations. Fixes #210

ekr:
Incorporate a revised version of @rohan-wire's text from Issue #210. Instead of telling people they should use encrypted messages, instead explain the consequences of doing so or not.

PR #247, opened 2024-01-30
Make encrypted group operation recommendation conditional.

Bren2010:


ekr:
As noted in the discussion about issue #246, I don't think this is the right answer.

draft-ietf-mls-architecture.md
```diff
@@ -915,8 +915,10 @@ While the Application messages will always be encrypted, having the handshake
 messages in plaintext has privacy consequences as someone could
 collect the signatures on the handshake messages and use them for tracking.
 
-> **RECOMMENDATION:** Prefer using encrypted group operation messages to avoid
-> privacy issues related to non-encrypted signatures.
+> **RECOMMENDATION:** Prefer using encrypted group operation messages unless a
+> Delivery Service explicitly needs plaintext group operations to provide
```
raphaelrobert, line 919:
```suggestion
> Delivery Service explicitly needs plaintext handshake messages to provide
```

rohanmahy (approved):
If we go with this PR, I prefer using "handshake" messages as Raphael suggests.

beurdouche (approved):

ekr:
Please back this out. I wasn't at the interim and there wasn't a consensus call on the list, so we shouldn't be landing anything until that happens.

seanturner:
Yeah we should back this out.  I'll set something up to finalize this.

Issue #252, opened 2024-03-23
Concurrency issue with reinitialization welcomes?


TWal:
[Section 11.2 of RFC 9420](https://www.rfc-editor.org/rfc/rfc9420.html#section-11.2) specify that reinitialization proceed in three phases, that may be performed by different members:
- send a ReInit proposal
- send a Commit that cover the ReInit proposal
- send Welcome messages to invite people to the new group

What would happen if two group members send Welcome messages in the third step concurrently? If I am not mistaken, the group would fork?

TWal:
(issue opened following discussions with @msprotz and @MartinSarkany)

Bren2010:
The Welcome message references a resumption PSK with the group id and epoch of the group being replaced. A strongly-consistent DS can tell you which Welcome came first, or with an eventually-consistent DS you would use some tie-breaker function. For all intents and purposes, this is the same process as deciding which Commit to accept.

TWal:
I agree that such constraints on the DS would solve this problem, however it seems to me that these constraints are not enforced by the document for Welcome messages, section 5.2 enforce these properties only on Commits but not on Welcomes.

ekr:
I don't have a well-formed opinion on the substantive question, but it's not clear to me that this is an issue on the architecture document, which is Informational and doesn't use 2119 language. If it's a problem we should fix 9420, I suspect.

@bifurcation ?

kkohbrok:
The reinit proposal includes the GroupID of the new group, so whoever creates the commit in the new group that adds all the members of the old group (presumably the creator of the new group). I don't think two members can create such a commit at the same time and thus send out competing welcomes.

Another thing that concerns me a little is that 9420 seems to require the new group id to match the old group id. I don't think that was intended, but I can't quite remember.

TWal:
To clarify, here is my concern explained with more details:
- say Alice commit a ReInit, this reinit contains a `group_id` (as mentioned by Konrad)
- say Bob and Charlie both want to send a Welcome message to the group members for the reinitialized group, a psk is used to transfer the old epoch secret to the new group
- if I understand things correctly, creating the new group state is not deterministic, although the `group_id` and psk are deterministic, other parameters are not, such as the new ratchet tree (what are the new leaf nodes, what is their position, etc)
- hence say that Bob's new group is different than Charlie's new group
- and say that Alice receives Bob's Welcome first, and Dave receives Charlie's Welcome first
[...]

PR #255, opened 2024-06-05
#252: Discuss Welcome/Commit coupling

Bren2010:


draft-ietf-mls-architecture.md
```diff
@@ -795,6 +795,36 @@ If the Commit references an unknown proposal, group members may need to solicit
...
+commits a ReInit proposal, this immediately freezes the existing group and
+triggers the creation of a new group with a new `group_id`.
+
+Ideally, the new group would be created by the same member that committed the
+`ReInit` proposal (including sending Welcome messages for the new group to all
+of the previous group's members). However this operation is not always atomic,
+so it's possible for a member to go offline after committing a ReInit proposal
+but before creating the new group. When this happens, it's necessary for another
+member to continue the reinitialization by creating the new group and sending
+out Welcome messages.
```
rohanmahy, line 820: Well, it is possible to design a client that does that, but it is also possible to design ones that always do generate the new group atomically. Is there a *technical* reason to NOT generate the ReInit Commit and create the new group as a single step?

Bren2010: There's not a technical reason to make it not atomic. Like you say, it's possible, and doing so would avoid this issue. But I don't think it's something worth requiring

rohanmahy (approved):

Issue #254, opened 2024-05-23
Confusing wording about key material sharing


istankovic:
[Here](https://github.com/mlswg/mls-architecture/blob/d5f7bb8a0cfa00fb1e815aa12629d848652c18b9/draft-ietf-mls-architecture.md?plain=1#L596-L599) we have:
```
The association between members' identities and signature keys is fairly
flexible in MLS.  As noted above, there is no requirement that all clients
belonging to a given user use the same key pair (in fact, such key reuse is
forbidden to ensure clients have independent cryptographic state).  A member can
```

However, in section 6.1, [we have](https://github.com/mlswg/mls-architecture/blob/d5f7bb8a0cfa00fb1e815aa12629d848652c18b9/draft-ietf-mls-architecture.md?plain=1#L819-L822):
```
most cases, a client corresponds to a specific device used by a user. If a user
has multiple devices, the user will generally be represented in a group by
multiple clients (although applications could choose to have devices share
keying material).  If an application wishes to implement operations at the level
```

It appears that the two paragraphs are conflicting each other (forbidden key reuse vs. sharing key material).
It would be good to clarify precisely which key material/keys are being referred to in both places to avoid confusion,
especially since there are recommendations further in the document to avoid key sharing.

seanturner:
Can I get a volunteer to draft up a PR? See [email](https://mailarchive.ietf.org/arch/msg/mls/UgfHJZKdvGAG6ETQWwfamQ254kM/).

PR #256, opened 2024-06-05
#254: Address key material sharing consistently

Bren2010:


draft-ietf-mls-architecture.md
```diff
@@ -1430,14 +1430,14 @@ deniability requires further analysis.
...
 ### Associating a User's Clients
 
-When the same user uses multiple clients, it may be possible for other members
-of a group to recognize all of those clients as belonging to the same user.  For
-example, all of a user's clients might present credentials authenticating the
-user's identity.  This association among devices might be considered a leak of
-private information.  The remainder of this section describes several approaches
-for addressing this.
+When a user has multiple devices, the base MLS protocol requires each device to
+be represented by a distinct client in the MLS groups that the user is a member
```
ekr, line 1434: I don't think this statement is correct, because as noted below it is possible to share state.

draft-ietf-mls-architecture.md
```diff
@@ -1430,14 +1430,14 @@ deniability requires further analysis.
...
+of. As a result, the other members of the group will be able to identify which
+of a user's devices sent each message, and therefore which device the user is
+currently using. For some applications, this may be an unacceptable breach of
+the user's privacy.
 
-This risk only arises when the leaf nodes for the clients in question provide
+Note that this risk only arises when the leaf nodes for the clients in question provide
```
ekr, line 1440: I don't think "Note that this" helps. I prefer the previous text.

draft-ietf-mls-architecture.md
```diff
@@ -1450,10 +1450,10 @@ the MLS group.  The application would need to provide a synchronization
 mechanism so that the clients' state remained consistent across changes to the
 MLS group.
 
->**RECOMMENDATION:** Avoid sharing cryptographic state between clients to
-> improve resilience against compromises.  An attacker could use one compromised
-> device to establish ownership of a state across other devices and reduce the
-> ability of the user to recover.
+>**RECOMMENDATION:** Only use secret state synchronization mechanisms that have
+> been widely reviewed and are known not to impair the security guarantees of
+> MLS. When this is not possible, avoid trying to synchronize state between
```
ekr, line 1455: I don't think this first sentence is an improvement.

ekr (commented):
The editorial changes at the beginning of this PR make sense. The substantive changes seem like regressions and I don't think we should make them. If you want to narrow the PR as indicated in the review I'll land it.

draft-ietf-mls-architecture.md
Bren2010: It's correct. Key word is "base protocol". It's possible, but you need to build a lot of extra machinery to make this secure

draft-ietf-mls-architecture.md
Bren2010: Multiple people have expressed need to share secret state, there are multiple drafts on the topic. You want to say "we recommend against this full stop" ?

draft-ietf-mls-architecture.md
ekr: These are different statements. The base protocol doesn't have any facilities for state synchronization. That does not mean that it *requires* each device to be represented by a distinct client, any more than the fact TLS doesn't have any facility for transporting keys and certificates requires each TLS server to have a distinct certificate.

draft-ietf-mls-architecture.md
```diff
@@ -1430,12 +1430,12 @@ deniability requires further analysis.
 
 ### Associating a User's Clients
 
-When the same user uses multiple clients, it may be possible for other members
-of a group to recognize all of those clients as belonging to the same user.  For
-example, all of a user's clients might present credentials authenticating the
-user's identity.  This association among devices might be considered a leak of
-private information.  The remainder of this section describes several approaches
-for addressing this.
+When a user has multiple devices, the base MLS protocol requires each device to
```
Bren2010, line 1433:
```suggestion
When a user has multiple devices, the base MLS protocol expects each device to
```
?

ekr: No, it simply has no opinion on the topic.

draft-ietf-mls-architecture.md
ekr: No, that's not correct. I don't have a strong opinion on key sharing one way or the other. Rather, this text went through WGLC and what it says is that you shouldn't share state because one device might be compromised and impair the ability of the user to recover. This threat is not addressed by having a secure way to share keys, and so the text you propose is a non sequitur.

ekr (changes requested):

rohanmahy (approved):

Issue #251, opened 2024-03-23
Can the Delivery Service actually filter Commit messages?


TWal:
The document says (5.2.1)

> The Delivery Service can rely on the epoch and content_type fields of an MLSMessage for providing an order only to handshake messages, and possibly even filter or reject redundant Commit messages proactively to prevent them from being broadcast

I think this may be a problem if a malformed Commit for epoch `n` is sent by one of the group members, and this Commit is selected by the Delivery Service:
- group members will reject this Commit and stay at epoch `n`
- the Delivery Service will think the group has moved to epoch `n+1` and reject further Commits that may be sent by other group members (that are still in epoch `n`)

TWal:
(issue opened following discussions with @msprotz and @MartinSarkany)

kkohbrok:
Yes, this is an issue that we won't be able to fix without something like a ZKP. Even with a PublicMessage, the sender of the commit could encrypt garbage to subsets of its copath instead of the correct path secrets. There's already some work on this which you can find here: https://hal.science/hal-03558760/document

If I read it correctly, it shows it's possible, even though the performance might not be what most deployments will be able to stomach.

rohanmahy:
> I think this may be a problem if a malformed Commit for epoch `n` is sent by one of the group members, and this Commit is selected by the Delivery Service:
> 
>     * group members will reject this Commit and stay at epoch `n`
> 
>     * the Delivery Service will think the group has moved to epoch `n+1` and reject further Commits that may be sent by other group members (that are still in epoch `n`)

There are three cases when this could happen:
a) the sending member has a bug that causes it to accidentally send an invalid Commit (or detached proposal)
b) a receiving member has a bug that causes it to wrongly treat as invalid a valid Commit or referenced proposal
c) the sending member maliciously sends an invalid Commit or Proposal that looks fine to the hub but invalid to the clients in order to desychronize the group from its hub

What remedies do we have?
- tell the hub that a received Commit was invalid and ask it to rollback to the previous epoch.  Unfortunately without more safeguards (like more evidence or a quorum), this would allow a member who was properly removed from the group to reverse their removal.
- a member receiving an "invalid-to-it" Commit could request the GroupInfo from the hub, rejoin via External Commit, then Commit to Remove the sender of the "invalid" Commit, and restore any changes. 

Note that cases a) and b) can occur even if the hub has no decision making in the group other than ordering.

TWal:
[...]

PR #257, opened 2024-06-05
#251: Advise against absolutist DS

Bren2010:


ekr:
I don't understand what this is trying to say, and it seems to contradict the text directly above.

Bren2010:
Adding further explanation. Blocking/filtering is fine, just not reliable, like you said, so DS should eventually defer to client

draft-ietf-mls-architecture.md
```diff
@@ -760,8 +760,14 @@ they don't receive another Commit first.
 The Delivery Service can rely on the `epoch` and `content_type` fields of an
 MLSMessage for providing an order only to handshake messages, and possibly even
 filter or reject redundant Commit messages proactively to prevent them from
-being broadcast. Alternatively, the Delivery Service could simply apply an order
-to all messages and rely on clients to ignore redundant Commits.
+being broadcast. However, this is not to say that the Delivery Service should
+enforce an absolutist view of the current group `epoch` and reject all messages
+that do not match it. Delivery Services have a significantly limited view of the
```
rohanmahy, line 765: "Delivery Services have a significantly limited view of the group state, and of which messages will be acceptable to clients"
This is often false with many implemented and deployed DSes.

**Some** DS *might have* ...

rohanmahy, line 765:
```suggestion
that do not match it. Some Delivery Services might have a limited view of the
```

draft-ietf-mls-architecture.md
```diff
@@ -760,8 +760,14 @@ they don't receive another Commit first.
...
+group state, and of which messages will be acceptable to clients. Once a
+Delivery Service is confident that a given commit was processed by a client, if
+the client continues to send messages in the previous epoch, the Delivery
+Service should assume the commit was silently rejected and stop
+blocking/filtering.
```
rohanmahy, line 770: This last sentence is speculative and very specific to a particular architectural choice. Please delete it.

```suggestion
group state, and of which messages will be acceptable to clients.
```

rohanmahy: If the purpose of the PR is to remove speculative language about what a DS might do, the cure here is worse than the disease and introduces further speculative text.

draft-ietf-mls-architecture.md
Bren2010: I'm not aware of a DS that can verify MACs, UpdatePaths, confirmation keys. That's what I'm referring to here

draft-ietf-mls-architecture.md
Bren2010: The purpose of this PR is to say that the DS may filter commits, in the interest of efficiency / etc, but can't be the ultimate arbiter of which commit is correct and needs to defer to clients eventually. The last sentence was added to make the deference point more explicit after ekr's comment: https://github.com/mlswg/mls-architecture/pull/257#issuecomment-2151042899

rohanmahy: But you are making a very specific comment about what the DS should assume which I believe is rarely correct and has substantial DoS attack consequences. It introduces a whole new attack surface.

Bren2010: The DoS attack is documented in the parent issue #251, i.e., when the DS tries to enforce an incorrect view of the epoch. The alternate approaches that you bring up in #251 are not secure:

- Asking the DS to rollback an invalid commit. As you say there, there's no way to prevent this from being abused. Proving to the DS that a commit is invalid would require revealing secret keys in a way that would impair group security. Trying to guard with a quorum is also vulnerable to sybil.
- Members do an external join. This still has DoS potential, and allows the DS / an attacker to make arbitrary changes to the ratchet tree & group state, which members will then accept uncritically.

The only way to mitigate DoS and preserve group security is for DS to defer to clients

ekr:
Taking a step back, the idea here is to deliver this document, not to litigate issues that are arising in MIMI. This text had WG consensus, but obviously people now believe it is imperfect, so let's ask what the minimum change we can do to make it acceptable. Given that the section is not even about filtering but rather about enforcing a *consistent* view of messages, I suggest we simply remove the text about filtering. I.e.,

> The Delivery Service can rely on the `epoch` and `content_type` fields of an
   MLSMessage for providing an order only to handshake messages. Alternatively,
  the Delivery Service could simply apply an order to all messages and rely on clients
  to ignore redundant Commits.

Bren2010:
I'm agreeable to that text. But I would also say: given that MIMI and several implementors have gotten this wrong, that makes it all the more important to find consensus, somewhere, if not here

kkohbrok:
I'm not sure about right and wrong. In some systems, having the DS determine whether a commit is valid or invalid is a choice you can make. Sure, that choice has certain trade-offs in terms of DoS attacks, but if that's okay in your thread model, it's a valid choice. I would say we go with @ekr's minimal approach for now and then discuss if we want an additional (sub-)section that further discusses rejection of invalid commits by the DS and what that entails.

draft-ietf-mls-architecture.md
rohanmahy: I'm going to flatly disagree with your last sentence "the only way to mitigate DoS" and your proposed solution in the original PR. Your proposed solution assumes that the clients act correctly when the only possible way they could have gotten into the situation in the first place was a client bug or a malicious client. Rolling back to an earlier commit may in fact be the attacker's intended form of service denial. 

Let's be clear here that the DoS attack you are talking about is in practice limited to members (because in many cases the DS/AS knows the client credential that matches that member). It is completely reasonable in some environments that a DS "fix" a misbehaving client, by banishing the offending client, not by letting the client further run afoul of policy rules.

rohanmahy:
> I'm agreeable to that text. But I would also say: given that MIMI and several implementors have gotten this wrong, that makes it all the more important to find consensus, somewhere, if not here

Given? Wrong by what measure?  Because for every case where you've said the DS shouldn't filter because bad thing X happens, I can show a failure mode where bad thing X happens only with clients. At least three separate implementations independently came up with DS filtering and have been using it in non-trivial environments with large numbers of groups and clients. If you wish to share your operational experience showing superior results in a variety of situations (not just theoretically) then please do. I think you have some good ideas and good insights into client-oriented approaches, some of which may bear fruit for the entire community. However, saying that everyone else "got it wrong" without evidence that your proposal solves the overall issue any better is no way to move us to consensus.

draft-ietf-mls-architecture.md
Bren2010: I think you're confused about what I'm proposing here. Deference means that the DS allows clients to behave in ways that are potentially incorrect, and allowing honest clients to decide to ignore messages from misbehaving clients. Is there a specific example you could walk through, where deference doesn't work?

Yes, the DoS attack is limited to members. But it's not correct to say that the DS has any ability to identify which client was misbehaving. The DS isn't able to identify incorrect MACs, UpdatePaths, etc. You could try guarding this mechanism with a quorum vote, but again, that's vulnerable to sybil attacks.

Bren2010:
> Given? Wrong by what measure?

If an implementer made a DS that says "as soon as I see a commit with epoch+1, I'm going to force every message to be epoch+1", then that allows the group to be bricked by any member sending a commit that can't be processed. If an implementer wanted to prevent the group from being able to be bricked like that, and implemented the "external rejoin" idea discussed above, then their groups can still be bricked by a single member, and they've also undermined the security of their MLS implementation. So it's wrong in the sense that it doesn't achieve the intended goal, and potentially reduces security.

I also think you might be conflating two types of filtering. The DS /can/ inspect public handshake messages and messages that contain proposals that violate policy. This is what I think most implementors truly care about being able to do. However, the DS /should not/ try to force a given commit to be accepted.

> However, saying that everyone else "got it wrong" without evidence that your proposal solves the overall issue any better is no way to move us to consensus.

I have been making this point in various forums for months, in person, in writing, in presentations. The fact that a majority of people have done something in a certain way isn't an argument for it being secure. The MLS wg was created because secure messaging is very hard, and someone working in isolation will often implement it in a way that's subtly insecure.

bifurcation:
I agree that this PR does not match where we ended up with in the issue discussion.  It is more prescriptive, and make more assumptions about the internal structure of the DS.  I have written up an alternative in #258

Bren2010:
What assumption am I making about the internal structure of the DS?

PR #258, opened 2024-06-10
Describe the risks of filtering at the DS

### LINKED ISSUE #251: Can the Delivery Service actually filter Commit messages?

bifurcation:
Fixes #251

### DIFF
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -763,6 +763,28 @@ filter or reject redundant Commit messages proactively to prevent them from
 being broadcast. Alternatively, the Delivery Service could simply apply an order
 to all messages and rely on clients to ignore redundant Commits.
 
+There is some risk associated with filtering.  Situations can arise where a
+malicious or buggy client sends a Commit that is not accepted by some members of
+the group, and the DS is not able to detect this and reject the Commit.  For
+example, a buggy client might send a encrypted Commit with an invalid set of
+proposals.  Or a malicious client might send a malformed Commit of the form
+described in {{Section 16.12 of RFC9420}}.
+
+In such situations, the DS might update its internal state under the assumption
+that the Commit has succeeded and thus end up in a state inconsistent with the
+members of the group.  For example, the DS might think that the current epoch is
+now `n+1` and reject any commits from other epochs, while the members think the
+epoch is `n`, and as a result, the group is stuck -- no member can send a Commit
+that the DS will accept.
+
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-architecture.md
```diff
@@ -763,6 +763,28 @@ filter or reject redundant Commit messages proactively to prevent them from
...
+epoch is `n`, and as a result, the group is stuck -- no member can send a Commit
+that the DS will accept.
+
+Given these risks, it is effectively impossible for a strongly consistent DS to
+know with absolute certainty when it is safe to update its internal state.
+Nonetheless, there are some simple mechanisms that can mitigate these risks.
+For example, having the DS wait to update its state until it gets confirmation
+of acceptance from a quorum of members provides some protection against buggy
+clients.  It is up to the designers and operators of a DS to ensure that
+sufficient mechanisms are in place to address these risks.
```
Bren2010, line 786: I believe this PR is essentially the same as mine except that you've removed the proposal to defer to clients / allow forks, with a proposal to have a quorum. Is that a correct statement? Why do you think quorum is better than allowing forks, when quorum requires all group members to be online and forking can be offline?

TWal:
I think this PR does a good job at framing the problem described in #251, however I think it would be best to refrain from trying to give solutions: the only solution would be to transmit all commits in a consistent order, and any attempt to filter may lead to DoS. The architecture document should allow the DS to filter (because actual deployments will filter), but warn about the risks it implies.

ekr:
I propose we take Richard's first two grafs more or less as-is and then reduce the last graf to say:

> Given these risks, it is effectively impossible for a strongly consistent DS to
know with absolute certainty when it is safe to update its internal state.
It is up to the designers and operators of a DS to ensure that
sufficient mechanisms are in place to address these risks.

bifurcation:
@ekr done

ekr (approved):
LGTM. @rohanmahy  any objections?

rohanmahy (approved):
works for me

PR #260, opened 2024-06-13
fix figure 2

stsch9:
Hi,

I think there is a small mistake in [Figure 2](https://messaginglayersecurity.rocks/mls-architecture/draft-ietf-mls-architecture.html#figure-2), as it says in section [3.3](https://messaginglayersecurity.rocks/mls-architecture/draft-ietf-mls-architecture.html#section-3.3):

> When Alice wants to create a group including Bob, she first uses the DS to look up his initial keying material. 

The same is the case for section [3.4](https://messaginglayersecurity.rocks/mls-architecture/draft-ietf-mls-architecture.html#section-3.4)


bifurcation (approved):
This seems right to me.

PR #264, opened 2024-07-27
Address comments from Yoav Nir and Rohan Mahy. Fixes #263

### LINKED ISSUE #263: Replace unusual meaning of verb "privilege"
"Otherwise privilege a stronger separation."
This uses a rare meaning of the word. Remove this sentence or say this in a different way.

ekr:


draft-ietf-mls-architecture.md
```diff
@@ -1418,12 +1418,13 @@ MLS provides additional protection regarding secrecy of past messages and future
...
-means that clients have the extremely important role of deleting appropriate
-keys as soon as they have been used with the expected message, otherwise the
-secrecy of the messages and the security for MLS is considerably weakened.
+FS means that access to all encrypted traffic history combined with
+access to all current keying material on clients will not defeat the
+secrecy properties of messages older than the oldest key of the
+compromised client.  Note that this means that clients the appropriate
+keys as soon as they have been used with the expected message,
+otherwise the secrecy of the messages and the security for MLS is
+considerably weakened.
```
rohanmahy, line 1427:
```suggestion
compromised client.  Note that this means that clients have to delete the appropriate
keys as soon as they have been used with the expected message,
otherwise the secrecy of the messages and the security for MLS is
considerably weakened.
```

ekr: Argh. thanks for the fix.

PR #269, opened 2024-11-05
Add recommendation for deletion of `init_key`s

TWal:
Because the `joiner_secret` is encrypted to the `init_key` of joiners, if the joiners don't delete their `init_key` after processing a Welcome, this could undermine forward-secrecy. I noticed that the document don't give any precise recommendations about that.

There are some hints scattered in the document, but they give recommendations to participants adding other participants, not to participants being added: https://github.com/mlswg/mls-architecture/blob/b091b3a30d6f9d2f46383f7bd152058f0fbfc9b1/draft-ietf-mls-architecture.md?plain=1#L670-L672

This PR adds a recommendation for that. I am not sure on how to proceed, I have made an attempt but it might belong to another section!

### DIFF
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -669,12 +669,13 @@ multiple groups before needing to upload more fresh KeyPackages.
 
 In order to avoid replay attacks and provide forward secrecy for messages sent
 using the initial keying material, KeyPackages are intended to be used only
-once. The Delivery Service is responsible for ensuring that each KeyPackage is
-only used to add its client to a single group, with the possible exception of a
-"last resort" KeyPackage that is specially designated by the client to be used
-multiple times. Clients are responsible for providing new KeyPackages as
-necessary in order to minimize the chance that the "last resort" KeyPackage will
-be used.
+once, and `init_key` is intended to be deleted by the client after decryption
+of the Welcome message. The Delivery Service is responsible for ensuring that
+each KeyPackage is only used to add its client to a single group, with the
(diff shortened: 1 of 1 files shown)


TWal:
Thinking more about init keys, we should also recommend to regularly rotate key packages that are on the delivery service, even if they are not used. This could otherwise undermine post-compromise security: the attacker could compromise the initialization key of someone, and when the corresponding key package is added in a group later (e.g. several months after the compromise) the attacker can decrypt the messages exchanged in this group, until the participant updates.

TWal:
There are some hints that key packages must expire: https://github.com/mlswg/mls-architecture/blob/b8893e0e158defae99ed57caca300a4151598fe0/draft-ietf-mls-architecture.md?plain=1#L1819-L1822
but I think kind of attack undermines post-compromise security rather than forward secrecy.

beurdouche (approved):

TWal:
I am wondering, should it be a "RECOMMENDATION"? It seems that other "RECOMMENDATION" are about things that can only be done in a best-effort way, whereas deleting initialization key upon Welcome processing seems to always be doable.

TWal:
I opened #274 to discuss KeyPackage expiration.

rohanmahy (approved):

seanturner:
In WGLC; see [this](https://mailarchive.ietf.org/arch/msg/mls/iDUWibpXgp2JT5dY2CkczzD4bmc/).

seanturner:

ekr:
Merged manually.

PR #270, opened 2024-11-05
Update github action

TWal:
The last two pull-requests failed because the version of `actions/upload-artifact` is too old (v2 instead of v4).
I have imported the new version of the action from [here](https://github.com/martinthomson/i-d-template/blob/main/template/.github/workflows/ghpages.yml) because the current `ghpages` workflow is adapted from [an old version (oct 2020) of it](https://github.com/martinthomson/i-d-template/blob/aa32825d42c99ef0eb7d40513379396f5af561ee/template/.github/workflows/ghpages.yml).

### DIFF
--- a/.github/workflows/ghpages.yml
+++ b/.github/workflows/ghpages.yml
@@ -20,40 +20,39 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: "Checkout"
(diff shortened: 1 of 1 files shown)


beurdouche:
Let's try this! : ) Thanks Théophile!

PR #274, opened 2025-02-27
Add recommendation about KeyPackage expiration

TWal:
As discussed in #269, KeyPackages must expire to have post-compromise security. Otherwise, if a participant generates a KeyPackage, the attacker compromises the initialization key, and the KeyPackage is used to invite the participant later (say, a year later), the attacker will know the group's epoch secrets despite having compromised the KeyPackage a while ago.

There are hints about KeyPackage expiration in the document, but it is not mentioned explicitly in the section about KeyPackages, hence this PR.

### DIFF
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -676,11 +676,6 @@ multiple times. Clients are responsible for providing new KeyPackages as
 necessary in order to minimize the chance that the "last resort" KeyPackage will
 be used.
 
-To provide post-compromise security for messages sent using the initial keying
-material, KeyPackages are intended to be rotated regularly, and expire using
-the `lifetime` field of the enclosed LeafNode, thereby ensuring old KeyPackages
-are not used to join a group.
-
 > **RECOMMENDATION:** Ensure that "last resort" KeyPackages don't get used by
 > provisioning enough standard KeyPackages.
 
@@ -690,8 +685,6 @@ are not used to join a group.
 
 > **RECOMMENDATION:** Ensure that the client for which a last resort KeyPackage
 > has been used is updating leaf keys as early as possible.
-
-> **RECOMMENDATION:** Enforce KeyPackage expiration using the `lifetime` field.
 
 Overall, it needs to be noted that key packages need to be updated when
 signature keys are changed.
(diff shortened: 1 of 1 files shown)


draft-ietf-mls-architecture.md
```diff
@@ -1818,8 +1825,8 @@ service matter, not via technology.
 
 Because the DS is responsible for providing the initial keying material to
 clients, it can provide stale keys. This does not inherently lead to compromise
-of the message stream, but does allow it to attack forward security to a limited
-extent. This threat can be mitigated by having initial keys expire.
+of the message stream, but does allow it to attack post-compromise security to
+a limited extent. This threat can be mitigated by having initial keys expire.
```
Bren2010, line 1829: I don't think that we need the text above, but this fix fs => pcs is a good catch

TWal: Maybe the text above is not in the correct place, but I feel like something needs to be said. The only place where the document says KeyPackage must expire (through the `lifetime` field) is here, where it only talks about the DS replaying old KeyPackages. But even with a honest DS, KeyPackages must expire for PCS.

Bren2010: The enforcement of the lifetime field is specified here: https://www.rfc-editor.org/rfc/rfc9420.html#section-7.3-4.5.1
So the text above seems duplicate, no?

TWal: You are right, the text above is duplicate with the mls-protocol rfc.

I think the text should be something saying that short lifetimes is better for PCS (and longer lifetimes is better for availability or something). Right now the mls-architecture document is not really talking about the implications of lifetimes, or is too vague about it (e.g. only talking about evil DS replaying old KeyPackages).

I don't know where such a text should belong though, I am not highly familiar with the mls-architecture rfc. What do you think about it?

Bren2010: That concern would seem to already be addressed by the paragraph we're commenting on: "the DS can provide stale keys; the solution is to set an expiration"

Also here: https://www.rfc-editor.org/rfc/rfc9420.html#section-16.6-5

TWal: Ok, this paragraph is exactly about what I wanted to mention in this document. I removed the text above.

Bren2010 (approved):

bifurcation (approved):

seanturner:
In WGLC, see [this](https://mailarchive.ietf.org/arch/msg/mls/ow9iTkmAZGrc7YmHp6AGpbatCMY/).

seanturner:

ekr:
Merged manually.

PR #313, opened 2025-03-04
Cite a new analysis of TreeKEM

TWal:
We recently put on eprint a formal analysis of TreeKEM in the symbolic model. This PR adds a citation for this paper.

### DIFF
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -205,6 +205,16 @@ informative:
       - name: Karthikeyan Bhargavan
     target: https://eprint.iacr.org/2025/410.pdf
 
+  CGWZ25:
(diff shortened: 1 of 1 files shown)


TWal:
I further added the recent paper of Cremers et al. on ETK.

I noticed that the text is somewhat imprecise, for example it says many of the papers "analyze the ratcheting tree sub-protocol of MLS" when the papers often analyze more than that (e.g. they also analyze the key schedule).

I can try to write a better text if you think that would be useful.

bifurcation (approved):

seanturner:
In WGLC, see [this](https://mailarchive.ietf.org/arch/msg/mls/ow9iTkmAZGrc7YmHp6AGpbatCMY/).

seanturner:

ekr:
Merged separately.

PR #315, opened 2025-03-11
Fixed whitespace in Group Formation Example

Qubitza:
While studying the document I found a minor whitespace issue.

A missing white space lead to a rendering issues in the HTML and PDF view (see https://datatracker.ietf.org/doc/html/draft-ietf-mls-architecture-15#section-3-2.1.1).


ekr:
Separately fixed by the RPC

PR #326, opened 2025-03-24
Use AS and DS most of the time. Fixes #279

### LINKED ISSUE #279: [rfced] Sections 2.2 and subsequent:  After the initial
[rfced] Sections 2.2 and subsequent:  After the initial
definition of "AS" in Section 2.2, we see alternating use of
"AS" and "Authentication Service".  Would you like to use the
abbreviation in running text after the initial definition?

We have the same question for
* "DS" and "Delivery Service"
* "FS" and "Forward Secrecy"
* "PCS" and "Post-Compromise Security"

It seems that each of these terms is used often enough that the
abbreviations could be used after the terms are first defined.

A few examples (for "DS" and "AS"; "Delivery Services, which is
  responsible ..." has been corrected):
 *  A Delivery Service (DS) which can receive and distribute messages
    between group members.  In the case of group messaging, the
    delivery service may also be responsible for acting as a
    "broadcaster" where the sender sends a single message which is
    then forwarded to each recipient in the group by the DS.
...
 It is important to note that the Authentication Service can be
 completely abstract in the case of a Service Provider which allows
 MLS clients to generate, distribute, and validate credentials
 themselves.  As with the AS, the Delivery Service can be completely
 abstract if users are able to distribute credentials and messages
 without relying on a central Delivery Service (as in a peer-to-peer
 system).  Note, though, that in such scenarios, clients will need to
 implement logic that assures the delivery properties required of the
 DS (see Section 5.2).
...
 She sends both of these messages to the Delivery Services, which is
 responsible for sending them to the appropriate people.  Note that
 the security of MLS does not depend on the DS forwarding the Welcome
 message only to Bob, as it is encrypted for him; it is simply not
 necessary for other group members to receive it.
...
 The Authentication Service design is left to the infrastructure
 designers.  In most designs, a compromised AS is a serious matter, as
 the AS can serve incorrect or attacker-provided identities to
 clients.

ekr:
I use the abbreviations for AS and DS most of the time after first use, with the major exception of heds and the like.
I also lower-cased Forward Secrecy and Post-Compromise Security

I haven't updated the XML yet, but if this looks good, I will harmonize after.

### DIFF
--- a/draft-ietf-mls-architecture.md
+++ b/draft-ietf-mls-architecture.md
@@ -244,7 +244,7 @@ informative:
 The Messaging Layer Security (MLS) protocol (RFC 9420)
 provides a group key agreement protocol for messaging applications.
 MLS is meant to protect against eavesdropping, tampering, and message
-forgery, and to provide Forward Secrecy (FS) and Post-Compromise Security
+forgery, and to provide forward secrecy (FS) and post-compromise security
 (PCS).
 
 This document describes the architecture for using MLS in a general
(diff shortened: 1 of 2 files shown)


bifurcation (approved):
No parallel XML changes?

ekr:
> I haven't updated the XML yet, but if this looks good, I will harmonize after. 

I will do that now.

PR #331, opened 2025-03-26
Minor edits from top-to-bottom readthrough

ekr:


ekr:
As usual, I'll fix up the XML after.

draft-ietf-mls-architecture.md
```diff
@@ -446,7 +447,7 @@ for the first time. This keying material is authenticated with their
 long-term credentials. Although in principle this keying material
 can be reused for multiple senders, in order to provide forward secrecy
 it is better for this material to be regularly refreshed so that each
-sender can use a new key.
+sender can use a new key, and older keys deleted.
```
bifurcation, line 450:
```suggestion
sender can use a new key and delete older keys.
```

I think this technically parses as-is, but seems awkward.

draft-ietf-mls-architecture.md
```diff
@@ -669,7 +672,7 @@ versions and ciphersuites. As such, there may be multiple KeyPackages stored by
 each user for a mix of protocol versions, ciphersuites, and end-user devices.
 
 When a client wishes to establish a group or add clients to a group, it first
-contacts the DS to request KeyPackages for each other client,
+contacts the DS to request KeyPackages for the other clients,
```
bifurcation, line 675:
```suggestion
contacts the DS to request a KeyPackage for each other client,
```

Emphasizes that you need a KeyPackage for each other client.

draft-ietf-mls-architecture.md
```diff
@@ -1137,7 +1140,7 @@ future. Thus, MLS offers a version negotiation mechanism; this mechanism
 prevents version downgrade attacks where an attacker would actively rewrite
 messages with a lower protocol version than the messages originally offered by the
 endpoints. When multiple versions of MLS are available, the negotiation protocol
-guarantees that the version agreed upon will be the highest version supported in
+guarantees that the version agreed upon will be the most preferred version supported in
```
bifurcation, line 1143: "Most preferred" seems to raise the question of "preferred by whom".  (The answer is "the creator", but might not want to get into that here.)  Maybe just "best version"?

draft-ietf-mls-architecture.md
```diff
@@ -1276,13 +1279,11 @@ deployments for them to interoperate:
...
-  signature keys, because each credential corresponds to a different client
-  owned by the same application user. However, one device may control multiple
-  signature keys -- for instance, if the device has keys corresponding to multiple
-  overlapping time periods -- but should still only be considered a single
-  client.
+- A policy for when two credentials represent the same client, distinguishing
+  the following two cases:
+  - When there are multiple devices for a given user.
+  - When a single has signature keys -- for instance, if the device has keys corresponding to multiple
+    overlapping time periods.
```
bifurcation, line 1286:
```suggestion
  - When a single device has multiple signature keys -- for instance, if the device has keys corresponding to multiple
    overlapping time periods.
```

draft-ietf-mls-architecture.md
```diff
@@ -1524,7 +1528,7 @@ properties, but defining the specific requirements and resulting notions of
 deniability requires further analysis.
 
 
-### Associating a User's Clients
+### Associating a User's Clients {#associating-a-users-clients}
```
bifurcation, line 1522: This tag shouldn't be necessary; kramdown auto-generates tags of this form based on the text of the heading.

bifurcation (approved):

draft-ietf-mls-architecture.md
ekr, line 675:
```suggestion
contacts the DS to request KeyPackages for each of the other clients,
```

ekr: I used "each of the other clients"

draft-ietf-mls-architecture.md
ekr: oops.

draft-ietf-mls-architecture.md
ekr: I was worried about the ' but yeah, it seems fine.

draft-ietf-mls-architecture.md
ekr: I don't think it guarantees that they will get the "best" version. For instance, there might be disagreement about best.  How about
"The negotiation protocol guarantees that the creator is able to select the best version out of..."

bifurcation: That works for me.

